PR #9, opened 2015-08-13
GHC 7.10.2 was released a few days ago

tebello-thejane:
Yes, I know that the build will fail with the error below. But that's the whole point of CI.

```
Math/NumberTheory/Moduli.hs:246:11:
    Not in scope: type constructor or class ‘Word64’
    Perhaps you meant ‘Word’ (imported from Prelude)
```


tebello-thejane:
Hmm.. The travis build succeeds? This is surprising.

On my system it fails to build on 7.10.2 (via ``cabal`` and via ``stack`` LTS-3.0) with
```
Math/NumberTheory/Moduli.hs:246:11:
    Not in scope: type constructor or class ‘Word64’
    Perhaps you meant ‘Word’ (imported from Prelude)
```

I will create a pull request to fix this shortly...

Bodigrim:
Travis build succeeds, because it is 64bit. And the only usage of `Word64` in `Math.NumberTheory.Moduli` is guarded by `#if WORD_SIZE_IN_BITS == 32`. Let us continue this discussion in #10.

Sorry, I have already updated `.travis.yml` to use GHC 7.10.3.

PR #10, opened 2015-08-13
Arithmoi fails to build on GHC 7.10.2 due to Word64 not being imported.

tebello-thejane:
Attempt to fix the following error when compiling on GHC 7.10.2:

```
    Not in scope: type constructor or class ‘Word64’
    Perhaps you meant ‘Word’ (imported from Prelude)
```

### DIFF
--- a/Math/NumberTheory/Moduli.hs
+++ b/Math/NumberTheory/Moduli.hs
@@ -34,9 +34,9 @@ module Math.NumberTheory.Moduli
 
 #include "MachDeps.h"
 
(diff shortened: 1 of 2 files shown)


cartazio:
Thanks. I'll get on this soon. I've accumulated a bit of a backlog here :)

Bodigrim:
@tebello-thejane your fix causes warning `The import of ‘Data.Word’ is redundant` on 64bit architecture. Could you please wrap import into `#if WORD_SIZE_IN_BITS == 32 ... #endif`?

PR #14, opened 2016-01-04
Test and bug fixes for Math.NumberTheory.Powers

Bodigrim:
I have written some basic tests for `Math.NumberTheory.Powers` modules. They triggered two bugs, which I have fixed.

First bug (and the nasty one) was caused by `unP` subroutine from `Math.NumberTheory.Powers`. Since its test expression did not depend on its arguments, it could cycle forever. 

Second bug is concerned with integer roots of high degree (64 and more) from `Word`. In such case intermediate computations in `newtonK` involved numbers far beyond `maxBound`, which caused either division by 0, or incorrect answers. 

Here is Travis build: https://travis-ci.org/Bodigrim/arithmoi/builds/100183595


cartazio:
Sweet.  I'll try to make some time to review and then merge in the next
week.  Much appreciated, the current code base is quite a beast

Bodigrim:
Any news?

cartazio:
Later this evening.  Sorry for the slow flow up!
Also always safe to ping or lurk on the numerical Haskell IRC channel.

Math/NumberTheory/Powers/Cubes.hs
```diff
@@ -61,9 +61,9 @@ integerCubeRoot n
 --   that is, the largest integer @r@ such that @r^3 <= n@.
 --   The precondition @n >= 0@ is not checked.
 {-# RULES
-"integerCubeRoot'/Int"  integerCubeRoot' = cubeRootInt'
-"integerCubeRoot'/Word" integerCubeRoot' = cubeRootWord
-"integerCubeRoot'/Igr"  integerCubeRoot' = cubeRootIgr
+"integerCubeRoot'/Int"     integerCubeRoot' = cubeRootInt'
+"integerCubeRoot'/Word"    integerCubeRoot' = cubeRootWord
+"integerCubeRoot'/Integer" integerCubeRoot' = cubeRootIgr
```
cartazio: nice cleanup

Math/NumberTheory/Powers/Fourth.hs
```diff
@@ -50,9 +50,9 @@ integerFourthRoot n
 --   that is, the largest integer @r@ with @r^4 <= n@.
 --   The condition is /not/ checked.
 {-# RULES
-"integerFourthRoot'/Int"  integerFourthRoot' = biSqrtInt
-"integerFourthRoot'/Word" integerFourthRoot' = biSqrtWord
-"integerFourthRoot'/Igr"  integerFourthRoot' = biSqrtIgr
+"integerFourthRoot'/Int"     integerFourthRoot' = biSqrtInt
+"integerFourthRoot'/Word"    integerFourthRoot' = biSqrtWord
+"integerFourthRoot'/Integer" integerFourthRoot' = biSqrtIgr
```
cartazio: also nice

arithmoi.cabal
```diff
@@ -105,9 +105,21 @@ test-suite spec
...
   build-depends:        base
-                        ,hspec
-                        ,arithmoi
+                        , arithmoi
+                        , tasty
+                        , tasty-smallcheck
+                        , tasty-quickcheck
+                        , tasty-hunit
+                        , QuickCheck
+                        , smallcheck
```
cartazio: what lower/upper bounds on these versions should we do? It'd be nice to make sure that we *know* that the deps they are ..
it'd be nice to have *some* mechanism for reconstituting when this is buildable.
Like, at the very least workingout what bounds on quickcheck/small check are "definitely" valid, and *lowerbounds* on deps. 

also why did you switch the test suite code to tasty rather than hspec? I dont really care, but would like some extra color

test-suite/Math/NumberTheory/Powers/CubesTests.hs
```diff
@@ -0,0 +1,75 @@
...
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
+
+module Math.NumberTheory.Powers.CubesTests
+  ( testSuite
+  ) where
+
+import Test.Tasty
+import Test.SmallCheck.Series
+
+import Data.Functor.Identity
```
cartazio: this might be me being a bit paranoid, but because you're defining orphan instances for a pretty common type (Identity), might it make sense to newtype it and do GND for the shared stuff? or just define a new one?

likewise, where are Positive and Nonnegative defined?

cartazio:
did a first cut of looking through, looks good, but hope you don't mind answering those questions :)

cartazio:
ie: the fixes are definitely correct, wouldn't mind a teeny bit of comments explaining what those properties definitions mean, and some clarification on the Identity stuff etc, but thats mostly just me trying to make sure this is an AWESOME PR

arithmoi.cabal
Bodigrim: Just because I worked a lot with `tasty` and never with `hspec`. Also AFAIK `hspec` lacks support of `smallcheck`.

test-suite/Math/NumberTheory/Powers/CubesTests.hs
Bodigrim: Yeah, good idea to replace `Identity` with newtype. 

`Positive` and `NonNegative` are from `Test.SmallCheck.Series`. That is why we need to write orphan instances to quickcheck them.

Bodigrim:
Thanks for review. Updated, please take a look again.

Bodigrim:
@cartazio do you receive my mails?

cartazio:
Yes. Just got back from 3 days of travel and currently snowed in

The best way to ping me asynchronously is the numerical Haskell IRC
channel.  I'll finish my review soon.

Bodigrim:
Sorry, I do not use IRC.

cartazio:
ok woops :) 

looks solid! i'll merge in now. a cursory google indicates hspec has a small check integration https://hackage.haskell.org/package/hspec-smallcheck 

if you prefer tasty thats fine too, we can flip flop on that later, i think you want this merged in :)

Issue #7, opened 2015-06-05
sieveFactor (factorSieve 2048) (29*73) == [(2117,1)]


CarlEdman:
This may just be a documentation bug or perhaps I am misreading this:

> sieveFactor fs n finds the prime factorisation of n using the FactorSieve fs. For negative n, a factor of -1 is included with multiplicity 1. After stripping any present factors 2, the remaining cofactor c (if larger than 1) is factorised with fs. This is most efficient of course if c does not exceed the bound with which fs was constructed. If it does, trial division is performed until either the cofactor falls below the bound or the sieve is exhausted. In the latter case, the elliptic curve method is used to finish the factorisation.

I understood this to mean that the FactorSieve is used to speed up the factorisation, but if an occasional argument to (sieveFactor fs) is higher than the argument to factorSieve, the results would still be correct (if less sped-up by the sieve).

However, this appears not to be the case.  For example with arithmoi 0.4.1.2 on Haskell Platform 2014.2 (i.e., ghc 7.8.3), I get, for example: '''sieveFactor (factorSieve 2048) (29*73)==[(2117,1)]'''

A some checks seem to find this discrepancy only in cases where the smallest prime factor is 29.

CarlEdman:
Under 0.4.1.3 and ghc 7.10.2 still:
~~~
sieveFactor (factorSieve 2048) (29*73)==[(2117,1)]
~~~
Is this a real bug or am I misunderstanding something?

cartazio:
I'll have to investigate myself , a lot of arithmoi is buggy / under tested
and has performance engineering that isn't validated with bench marks.

cartazio:
If you can do any further sleuthing in the mean time that would be super
appreciated

CarlEdman:
Actually, I discovered arithmoi over a year ago and when I benchmarked it against my own code, arithmoi just blew it away.  So since then, I've been using arithmoi for my own number theoretical investigations almost every day and that above bug is the only one I've found so far.

But by chance, I was just thinking about ways to improve arithmoi and, beside contributing a bunch of other standard number theoretic functions I've coded on top of arithmoi and which may be of interest to others, the biggest performance win I can see would be parallelisation of the code.  Many of the big number theory codes are embarrassingly parallelizable and I often see real-world 10x speed improvements on by 12-processor workstation just by adding a few judicious `par`s and `pseq`s.

cartazio:
Cool! That's good to know, but we should still get benchmarks, especially
[...]

PR #17, opened 2016-02-16
Tests and bug fixes for Lucas, MoebiusInversion, Primes

### LINKED ISSUE #2: Segmentation fault with primeList and primeSieve
Reported by @treeowl ([dafis/arithmoi#16](https://bitbucket.org/dafis/arithmoi/issue/16/segmentation-fault-with-primelist-and)).

To reproduce:

```
$ cabal sandbox init
$ cabal install arithmoi-0.4.1.1 hspec-2.1.2 --ghc-options=-fasm
$ cat >Spec.hs <<'EOF'
```
```haskell
import Test.Hspec
import Test.QuickCheck
import Math.NumberTheory.Primes

primesSumWonk :: Int -> Int
primesSumWonk upto = sum . takeWhile (< upto) . map fromInteger . primeList $ primeSieve (toInteger upto)

primesSum :: Int -> Int
primesSum upto = sum . takeWhile (< upto) . map fromInteger $ primes

main = do
  hspec $
    describe "Primes.primesSumWonk" $
      it "works like primesSum" $
        property $ \n -> let upto = abs (n `rem` 45)
                         in primesSum upto == primesSumWonk upto
````
```
EOF
$ cabal exec runhaskell Spec.hs

Primes.primesSumWonk
Segmentation fault (core dumped)
```

@treeowl adds:

> FYI, if you do decide to attempt to fix this, you might look at making it use the `bitvec` package for mutable bit vectors instead of weaving that junk into the prime sieve in a way that leaves me utterly baffled about just what it's supposed to be doing. Another very pleasant alternative might be to just steal sieves from the `NumberSieves` package. I have no idea about how their performance compares, theoretically, but anything is better than code that's blatantly _broken._

Bodigrim:
1. Add new cabal flag `check-bounds`, which replaces all unsafe array functions with safe ones. Incredibly useful for debugging.

2. Write tests for `Math.NumberTheory.Lucas`. No bugs found!

3. Write tests for `Math.NumberTheory.MoebiusInversion`. Fix segfault, caused by index `2` out of bounds `(0, 1)`, when `n=0`. Add `moebius` function to the library (required for tests).

4. Fix #2 and #7, both caused by incorrect indexing of sieves. One segfault less!

If there is no objection, I am going to merge this pull request in two weeks.

### DIFF
--- a/Math/NumberTheory/Logarithms.hs
+++ b/Math/NumberTheory/Logarithms.hs
@@ -36,10 +36,10 @@ import GHC.Word (Word(..))      -- Moved to GHC.Types
 
 import Data.Bits
 import Data.Array.Unboxed
-import Data.Array.Base (unsafeAt)
 
 import Math.NumberTheory.Logarithms.Internal
 import Math.NumberTheory.Powers.Integer
+import Math.NumberTheory.Unsafe
 #if __GLASGOW_HASKELL__ < 707
 import Math.NumberTheory.Utils  (isTrue#)
 #endif
--- a/Math/NumberTheory/Moduli.hs
+++ b/Math/NumberTheory/Moduli.hs
@@ -39,7 +39,6 @@ import Data.Word
 #endif
 import Data.Bits
 import Data.Array.Unboxed
-import Data.Array.Base (unsafeAt)
 import Data.Maybe (fromJust)
 import Data.List (nub)
 import Control.Monad (foldM, liftM2)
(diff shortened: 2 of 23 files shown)


treeowl:
@Bodigrim, this is absolutely tremendous. Many thanks. Do you think you understand the Sieve of Eratosthenes code well enough to document what it's doing, how it's doing it, and (ideally) *why* it does it like that? This whole implementation smells disgustingly like a "numerical recipe".

Bodigrim:
I understand some basic machinery beyond, but there are so many magic numbers, abbreviations and premature optimisations!  Both issues above were shallow and did not require to dig deep. I'm writing tests to cover sieves thoroughly: at least we must be sure that this "numerical recipe" is valid.

cartazio:
as @Bodigrim  says, theres deep design issues with current arithmoi, but before we do proper design or performance engineering, making sure we have *correct* algorithms, while not redesigning the universe, is phase zero

PR #20, opened 2016-02-23
Add Gaussian Integer support

cfredric:
Adds basic support for Gaussian integers. I'm happy to write tests, but I was having some trouble figuring out how to use the test harness. (My implementation of this in https://github.com/cfredric/NumberTheory has some tests, which all pass.)

### DIFF
--- a/Math/NumberTheory/GaussianIntegers.hs
+++ b/Math/NumberTheory/GaussianIntegers.hs
@@ -18,6 +18,12 @@ module Math.NumberTheory.GaussianIntegers (
     imag,
     conjugate,
     magnitude,
+    divModG,
+    divG,
+    modG,
+    quotRemG,
+    quotG,
+    remG,
     (./),
     (.%),
     (.^),
@@ -64,6 +70,51 @@ instance Num GaussianInteger where
         | a == 0 && b == 0 = z               -- hole at origin
         | otherwise        = z ./ abs z
 
+quotRemG :: GaussianInteger -> GaussianInteger -> (GaussianInteger, GaussianInteger)
+quotRemG = divHelper quot
+
+quotG :: GaussianInteger -> GaussianInteger -> GaussianInteger
+n `quotG` d = q where (q,_) = quotRemG n d
+
+remG :: GaussianInteger -> GaussianInteger -> GaussianInteger
+n `remG`  d = r where (_,r) = quotRemG n d
+
+divModG :: GaussianInteger -> GaussianInteger -> (GaussianInteger, GaussianInteger)
+divModG = divHelper div
+
+divG :: GaussianInteger -> GaussianInteger -> GaussianInteger
+n `divG` d = q where (q,_) = divModG n d
+
+modG :: GaussianInteger -> GaussianInteger -> GaussianInteger
+n `modG` d = r where (_,r) = divModG n d
+
+-- "div" truncates toward -infinity, "quot" truncates toward 0, but we need
+-- something that truncates toward the nearest integer. I.e., we want to
+-- truncate with "round".
+divToNearest :: Integer -> Integer -> Integer
+divToNearest x y = round ((x % 1) / (y % 1))
+
+divModWithRoundG :: GaussianInteger -> GaussianInteger -> (GaussianInteger, GaussianInteger)
+divModWithRoundG = divHelper divToNearest
+
+-- |Divide one Gaussian integer by another, rounding each component toward the
+-- nearest integer (or to the even integer, if equidistant).
+(./) :: GaussianInteger -> GaussianInteger -> GaussianInteger
+n ./ d = q where (q, _) = divModWithRoundG n d
+
+-- |Compute the remainder of division, when dividing and rounding to the nearest integer.
+(.%) :: GaussianInteger -> GaussianInteger -> GaussianInteger
+n .% d = r where (_, r) = divModWithRoundG n d
+
+divHelper :: (Integer -> Integer -> Integer) -> GaussianInteger -> GaussianInteger -> (GaussianInteger, GaussianInteger)
+divHelper divide g h =
+    let nr :+ ni = g * conjugate h
+        denom = magnitude h
+        q = divide nr denom :+ divide ni denom
+        p = h * q
+    in (q, g - p)
+
+
 -- |Conjugate a Gaussian integer.
 conjugate :: GaussianInteger -> GaussianInteger
 conjugate (r :+ i) = r :+ (-i)
@@ -71,26 +122,6 @@ conjugate (r :+ i) = r :+ (-i)
 -- |The square of the magnitude of a Gaussian integer.
 magnitude :: GaussianInteger -> Integer
 magnitude (x :+ y) = x * x + y * y
-
--- "div" truncates toward -infinity, "quot" truncates toward 0, but we need
--- something that truncates toward the nearest integer. I.e., we want to
--- truncate with "round".
-divToNearest :: (Integral a, Integral b) => a -> a -> b
-divToNearest x y = round ((x % 1) / (y % 1))
-
--- |Divide one Gaussian integer by another.
-(./) :: GaussianInteger -> GaussianInteger -> GaussianInteger
-g ./ h =
-    let nr :+ ni = g * conjugate h
-        denom    = magnitude h
-    in divToNearest nr denom :+ divToNearest ni denom
-
--- |Compute the remainder when dividing one Gaussian integer by another.
-(.%) :: GaussianInteger -> GaussianInteger -> GaussianInteger
-g .% m =
-    let q = g ./ m
-        p = m * q
-    in g - p
 
 -- |Compute whether a given Gaussian integer is prime.
 isPrime :: GaussianInteger -> Bool
@@ -199,7 +230,7 @@ factorise g
 trialDivide :: GaussianInteger -> [GaussianInteger] -> [(GaussianInteger, Int)] -> (GaussianInteger, [(GaussianInteger, Int)])
 trialDivide g [] fs = (g, fs)
 trialDivide g (pf : pft) fs
-    | g .% pf == 0 =
+    | g `modG` pf == 0 =
         let (cnt, g') = countEvenDivisions g pf
         in trialDivide g' pft ((pf, cnt) : fs)
     | otherwise    = trialDivide g pft fs
@@ -212,5 +243,5 @@ countEvenDivisions g pf = helper g 0
     where
     helper :: GaussianInteger -> Int -> (Int, GaussianInteger)
     helper g' acc
-        | g' .% pf == 0 = helper (g' ./ pf) (1 + acc)
+        | g' `modG` pf == 0 = helper (g' ./ pf) (1 + acc)
         | otherwise     = (acc, g')
--- a/Math/NumberTheory/GaussianIntegers.hs
+++ b/Math/NumberTheory/GaussianIntegers.hs
@@ -20,7 +20,6 @@ module Math.NumberTheory.GaussianIntegers (
 ) where
 
 import Prelude hiding (gcd)
-import Data.Bits
 import Data.List  (genericLength)
 import Data.Ratio ((%))
 import qualified Math.NumberTheory.GCD as GCD
@@ -38,7 +37,7 @@ data GaussianInteger = Integer :+ Integer deriving (Ord, Eq)
 
 instance Show GaussianInteger where
     show (a :+ b) = show a ++ op ++ b' ++ "i"
-        where op = if b > 0 then "+" else "-"
+        where op = if b >= 0 then "+" else "-"
               b' = if abs b /= 1 then show (abs b) else ""
 
 -- |The real part of a Gaussian integer.
@@ -93,7 +92,7 @@ g .% m =
 isPrime :: GaussianInteger -> Bool
 isPrime = Testing.isPrime . magnitude
 
--- |An infinte list of the Gaussian primes. This list is in order of ascending magnitude.
+-- |An infinite list of the Gaussian primes. This list is in order of ascending magnitude.
 primes :: [GaussianInteger]
 primes = [ a' :+ b'
             | mag <- Sieve.primes
@@ -127,7 +126,7 @@ roots m
 order :: Integer -> Integer -> Integer
 order x m = head [ ord
                  | ord <- [1 .. genericLength $ units m]
-                 , Moduli.powerMod x ord m == (1 :: (Integral a, Num a, Bits a, Eq a) => a)
+                 , Moduli.powerMod x ord m == 1
                  ]
 
 -- |Find a Gaussian integer whose magnitude squared is the given prime number.
@@ -165,12 +164,12 @@ factorise g
                     in helper pt (g' ./ gp .^ pow) ((gp, pow) : fs)
                 | otherwise      =
                     -- general case: for every prime factor of the magnitude
-                    -- squared, find a gaussian prime whose magnitude squared
+                    -- squared, find a Gaussian prime whose magnitude squared
                     -- is that prime. Then find out how many times the original
-                    -- number is divisible by that gaussian prime and its
+                    -- number is divisible by that Gaussian prime and its
                     -- conjugate. The order that the prime factors are
                     -- processed doesn't really matter, but it is reversed so
-                    -- that the Guassian primes will be in order of increasing
+                    -- that the Gaussian primes will be in order of increasing
                     -- magnitude.
                     let gp = findPrime p
                         (!gNext, !facs) = trialDivide g' [gp, conjugate gp] []
--- a/Math/NumberTheory/GaussianIntegers.hs
+++ b/Math/NumberTheory/GaussianIntegers.hs
@@ -180,10 +180,10 @@ findPrime p
 a .^ e
     | e < 0 && norm a == 1 =
         case a of
-            1      -> 1
-            (-1)   -> if even e then 1 else (-1)
-            0 :+ 1 -> (0 :+ (-1)) .^ (abs e `mod` 4)
-            _      -> (0 :+ 1) .^ (abs e `mod` 4)
+            1    :+ 0 -> 1
+            (-1) :+ 0 -> if even e then 1 else (-1)
+            0    :+ 1 -> (0 :+ (-1)) .^ (abs e `mod` 4)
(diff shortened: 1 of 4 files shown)


Bodigrim:
Nice to see you! Thanks for PR!

Yeah, integrating with the test harness is not very easy. Here is a stub for your module. Please put it into `test-suite/Math/NumberTheory/GaussianIntegersTests.hs`. Feel free to modify :)

```haskell
module Math.NumberTheory.GaussianIntegersTests
  ( testSuite
  ) where

import Test.Tasty

import Math.NumberTheory.GaussianIntegers
import Math.NumberTheory.TestUtils

-- | Number is equal to the product of its factors.
factoriseProperty :: Integer -> Integer -> Bool
factoriseProperty x y = g == g'
  where
    g = x :+ y
    factors = factorise g
    g' = foldl (.*) (1 :+ 0) $ map (uncurry (.^)) factors

-- | Number is prime iff it is non-zero
--   and has exactly one (non-unit) factor.
isPrimeProperty :: Integer -> Integer -> Bool
isPrimeProperty x y
  =  x == 0 && y == 0
  || isPrime g && n == 1
  || not (isPrime g) && n /= 1
  where
    g = x :+ y
    factors = factorise g
    nonUnitFactors = filter (\(p, _) -> magnitude p /= 1) factors
    -- Count factors taking into account multiplicity
    n = sum $ map snd nonUnitFactors

testSuite :: TestTree
testSuite = testGroup "GaussianIntegers"
  [ testSmallAndQuick "factorise"   factoriseProperty
  , testSmallAndQuick "isPrime"     isPrimeProperty
  ]
```

Other changes:

```diff
--- a/arithmoi.cabal
+++ b/arithmoi.cabal
@@ -123,7 +124,8 @@ test-suite spec
                         , tasty-hunit >= 0.9 && < 0.10
                         , QuickCheck >= 2.8 && < 2.9
                         , smallcheck >= 1.1 && < 1.2
-  other-modules :   Math.NumberTheory.GCDTests
+  other-modules :   Math.NumberTheory.GaussianIntegersTests
+                  , Math.NumberTheory.GCDTests
                   , Math.NumberTheory.GCD.LowLevelTests
                   , Math.NumberTheory.LogarithmsTests
                   , Math.NumberTheory.LucasTests

--- a/test-suite/Test.hs
+++ b/test-suite/Test.hs
@@ -18,6 +18,8 @@ import qualified Math.NumberTheory.Powers.SquaresTests as Squares

 import qualified Math.NumberTheory.PrimesTests as Primes

+import qualified Math.NumberTheory.GaussianIntegersTests as Gaussian
+
 main :: IO ()
 main = defaultMain tests

@@ -47,4 +49,7 @@ tests = testGroup "All"
   , testGroup "Primes"
     [ Primes.testSuite
     ]
+  , testGroup "Gaussian"
+    [ Gaussian.testSuite
+    ]
   ]
```

If my explanations are vague, fill free to ask about any details here or [Join the chat at https://gitter.im/cartazio/arithmoi](https://gitter.im/cartazio/arithmoi?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)

Again many thanks for your contribution!

cartazio:
And there's also folks on the numerical Haskell IRC channel too ;)

cfredric:
I knew your username looked vaguely familiar! Thanks very much for the stub, and showing the rest of the changes I'd need to make.

I found a few typos and one other thing that I missed (an artifact from fighting with ghc earlier), added your test suggestions, and fixed the bug they revealed. Also added another test to make sure all the Gaussian integers in "primes" are actually prime. (Let me know if there's a way to tell QuickCheck and SmallCheck that that test requires a nonnegative Int - I wasn't sure, so I wrote around it.)

cfredric:
Following the missed prime case earlier, I just realized my definition for "primes" was incomplete as well. Fixed now. Sorry for not catching these sooner!

Bodigrim:
That's nice, thank you. 

I have several suggestions aimed to improve API. What is your opinion?

1. We can use record syntax to define `Gaussian` with accessors at once. E. g., `data GaussianInteger = (:+) { real :: !Integer, imag :: !Integer }`. Note bangs; strict fields usually decrease number of allocations and indirections, improving overall performance. One can come across many examples of bang patterns in `arithmoi`.

2. IMHO it is worth to make `Gaussian` an instance of `Num`. We'll be able to use usual arithmetic operators and numeric literals. Special operators `(.+)`, `(.-)`, `(.*)` tend to confuse users. You can define `abs` and `signum` as `error`, if you like. On the other hand, if you have enough free time, there is another (more fruitful) approach: define `abs z` as `z`'s associate from first quadrant. For instance, `abs (-1 :+ 2) == 2 :+ 1`. In this case `signum` is well-defined by equation `z = abs z * signum z`.

cfredric:
I think those suggestions make sense. (I wasn't aware you could use record syntax with an infix constructor, that's nice!) I've made both of those changes, as well as a few other minor updates - let me know what you think.

Bodigrim:
Wow, you are super fast! And it is cool that you keep adding tests.

Sorry, I am busy at the moment and cannot write all my suggestions at once. I hope you do not mind.

3. We cannot make `GaussianInteger` an instance of `Integral`, because the latter is a subclass of `Real` and `Enum`. But we still can follow its approach to API for modular computations. Instances of `Integral` define `quotRem` and get `quot`, `rem`, `div`, `mod`, `divMod` for free:

    ```haskell
    n `quot` d          =  q  where (q,_) = quotRem n d
    n `rem` d           =  r  where (_,r) = quotRem n d
    n `div` d           =  q  where (q,_) = divMod n d
    n `mod` d           =  r  where (_,r) = divMod n d

    divMod n d          =  if signum r == negate (signum d) then (q-1, r+d) else qr
                           where qr@(q,r) = quotRem n d
    ```

    I suggest to combine existing `(./)` and `(.%)` into gaussian analog of `quotRem` (`quotRemG` or `gQuotRem` or whatever you like) and define analogs of `quot`, `rem`, `div`, `mod`, `divMod` following (more or less) the snippet above. Gaussian `quotRem` should round towards `0 :+ 0`, and `divMod` should round towards `-Inf :+ -Inf`.

4. Could you please rename `gcd`? It is exported outside and will clash with `Prelude`.

5. (Debatable, feel free to skip.) I wonder whether it is possible to reuse `Math.NumberTheory.Moduli.sqrtModP` to `findPrime`. It would be nice to benchmark both variants: `arithmoi` is obsessed with performance.

6. Note that `isPrime` from `Math.NumberTheory.Primes` checks whether a number is a (negative or positive) prime element of the ring Z, but `primes` lists only positive primes. API for gaussian primes must be consistent with API for real ones. So I suggest that `Math.NumberTheory.GaussianIntegers.primes` should list only primes from the first quadrant. This will benefit both performance and practical applications.

What do you think?

cfredric:
I've had some free time the past few days, plus this has been a pet project of mine for a little while, so it's nice to see it going somewhere useful. I'll be out of town for a few days (starting tomorrow), so you'll have some time to put your suggestions down. :)

1. Providing versions of quotRem, divMod, etc., seems like a good idea. I didn't use the same definition of divMod in terms of quotRem since I wasn't sure if it would be quite valid in the Gaussian integers. However, I do think we need to keep the current implementations of `.%` and `./` in some form, in order to define `gcdG` properly. I had some issues with this when I was implementing `factorise` originally: if you define division and modulus of Gaussian integers using either `div` or `quot`, the obvious implementation of  `gcdG` doesn't terminate on all inputs. (`gcdG (12 :+ 23) (23 :+34)` is an example of this: the procedure terminates only when division rounds toward the nearest integer, not toward `-Inf` or toward `0`. I.e., division must round using `round`.)

Let me know what you think for new names for `.%` and `./`. I couldn't think of anything I liked, so I left them as is for now.

2. Done.

3. I'm not sure I see how the implementation of that would go - I might be forgetting some of my abstract algebra (it's been a while since I did much formal math!). Could you elaborate? 

4. That makes sense - consistency of API is definitely important, and performance can't be ignored either. Done.

Bodigrim:
Thanks for your patience :)

1. That's very interesting, never thought about an algorithm for  complex `gcd`. I'll investigate further in my spare time. For now could you please add `gcdG (12 :+ 23) (23 :+ 34)` as a `TestCase` (see `Math.NumberTheory.Powers.SquaresTests` for examples)? Do we need to export `.%` and `./` outside?

2. Thanks.

3. Here is a sketch of an algorithm. I am not sure that it is bullet-proof, would be nice to know your opinion. It is possible that this topic is covered by Bach and Shallit, but I do not have a copy at hand.

  Our aim is to solve Diophantine equation `a^2 + b^2 = p` for positive `a` and `b`. It implies that `a ≤ sqrt(p)`, `b ≤ sqrt(p)` and `a^2 ≡ -b^2 (mod p)`. The latter means that `a ≡ sqrt(-1) b (mod p)`, where `sqrt(-1)` stands for positive `c` such that `c^2 ≡ -1 (mod p)`.

  For example, take `p = 61`. We have `sqrtModP (-1) p == Just 11`, so `c = 11`. Now take `as = [1 .. integerSquareRoot 61]` (so `as = [1..7]`), build `bs` multiplying `as` by 11 mod 61 and filter them by predicate `b <= integerSquareRoot 61`. In our case we get `bs = [11,22,33,44,55,5,15]` and only `b=5` is less than `integerSquareRoot 61`. It corresponds to `a=6`. Right, `6^2 + 5^2 = 61`.

  The algorithm above seems to have O(sqrt p) complexity.

4. (Debatable, feel free to decline.) I am wondering: maybe it is worth to define `ι = 0 :+ 1`? `Math.NumberTheory.Factorisation` already defines `φ`, `λ`, `τ` and `σ`, so this will not be out of style. We can also use it for `Show` instance: e. g., show (1:+2) == '1+2*ι'.

5. Could you please fix the following GHC warning?

		Math/NumberTheory/GaussianIntegers.hs:186:9: Warning:
		    Pattern match(es) are overlapped
		    In a case alternative:
		        0 :+ 1 -> ...
		        0 :+ (-1) -> ...

		Math/NumberTheory/GaussianIntegers.hs:186:9: Warning:
		    Pattern match(es) are non-exhaustive
		    In a case alternative:
		        Patterns not matched: #x with #x `notElem` [1#, -1#]

cfredric:
No worries! :)

(In order to keep the numbering from piling up, I'll respond to your points in order, but not necessarily all of them.)

1. I was mulling it over on the way home today, and realized that if I made sure to keep the integers in the first quadrant at all stages of the Euclidean algorithm, it would probably avoid the rounding issue that `remG` and `modG` were having. Sure enough, that seems to work!

2. Thanks for that explanation! I never actually had any textbook for Number Theory when I took it during undergrad (the professor just wrote everything on the board, and we worked from that), and I have since lost my notes, so I forgot that Gaussian integers were related to that problem. I'll look into implementing that with `sqrtModP`, and see about testing and benchmarking it. I haven't looked at any of the benchmarking of arithmoi yet, though.

3. Sure, I have no objections to defining that as a constant. I may not use it internally though, just because I'd have to copy and paste it everywhere rather than being able to type it "by hand".

4. I fixed the second warning. (The non-exhaustive pattern - which, by the way, doesn't show up when I build. What version of GHC are you using?) But I can't figure out why the first warning is coming up. It seems to me that `1`, `(-1)`, and `(1 :+ 0)` are all disjoint cases, so I'm not sure why GHC disagrees. Do you have any thoughts on it?

Bodigrim:
1. Wow, great.

2. I have several benchmarks in my private branches, but none of them was merged into `master`. Here is a template, you can put it into `benchmark/Bench.hs`.

    ```haskell
    import Criterion.Main

    alternative1 x = ...

    alternative2 x = ...

    benchAlternatives lim = bgroup (show lim)
      [ bench "1" $ nf (map alternative1) [1..lim]
      [ bench "2" $ nf (map alternative2) [1..lim]
      ]

    main = defaultMain
        [ benchAlternatives 10
        , benchAlternatives 100
        , benchAlternatives 1000
        , benchAlternatives 10000
        , benchAlternatives 100000
        , benchAlternatives 1000000
        ]
    ``` 

3. Thanks.

4. It is GHC 7.10. I do not understand, why patterns are overlapped, too. I managed to fix the warning with
    
    ```haskell
    case a of
        1 :+ 0    -> 1
        (-1) :+ 0 -> if even e then 1 else (-1)
        0 :+ 1    -> (0 :+ (-1)) .^ (abs e `mod` 4)
        _         -> (0 :+ 1) .^ (abs e `mod` 4)
    ```

5. Could you please move `LANGUAGE` pragma under haddock comment?

cfredric:
1. I benchmarked the two alternatives, and your suggestion is unquestionably the way to go. I didn't even let the benchmark run to completion since it hadn't finished after two hours, but the last part that ran for both variants produced the following:

    benchmarking 100000/Via GCD
    time                 257.4 s    (256.2 s .. 258.0 s)
                         1.000 R²   (1.000 R² .. 1.000 R²)
    mean                 258.0 s    (258.0 s .. 258.0 s)
    std dev              54.35 ms   (0.0 s .. 55.25 ms)
    variance introduced by outliers: 19% (moderately inflated)
    
    benchmarking 100000/Via Sqrt
    time                 45.50 ms   (45.47 ms .. 45.52 ms)
                         1.000 R²   (1.000 R² .. 1.000 R²)
    mean                 45.48 ms   (45.47 ms .. 45.50 ms)
    std dev              26.51 μs   (19.82 μs .. 36.00 μs)

2. Done.
3. Done.

By the way, thank you for your patience with all of these iterations, and thank you for your suggestions! I appreciate that you have a high standard for this codebase.

cfredric:
Just rebased on master to get rid of the conflicts. Apologies if I didn't do that the right way (I see that the commits and comments are no longer interspersed).

Math/NumberTheory/GaussianIntegers.hs
```diff
@@ -0,0 +1,233 @@
...
+import qualified Math.NumberTheory.Primes.Factorisation as Factorisation
+import qualified Math.NumberTheory.Primes.Sieve as Sieve
+import qualified Math.NumberTheory.Primes.Testing as Testing
+
+infix 6 :+
+infixr 8 .^
+-- |A Gaussian integer is a+bi, where a and b are both integers.
+data GaussianInteger = (:+) { real :: !Integer, imag :: !Integer } deriving (Eq)
+
+ι :: GaussianInteger
```
Bodigrim: Could you please add haddock comment for `ι`?

Math/NumberTheory/GaussianIntegers.hs
```diff
@@ -0,0 +1,233 @@
...
+ι = 0 :+ 1
+
+instance Show GaussianInteger where
+    show (a :+ b) = show a ++ op ++ b' ++ "ι"
+        where op = if b >= 0 then "+" else "-"
+              b' = if abs b /= 1 then show (abs b) else ""
```
Bodigrim: (Debatable.) IMHO it is worth to change `Show` instance to produce 
`[1+ι,3,1+2*ι,2+ι,7,11,2+3*ι,3+2*ι]` instead of `[1+ι,3+0ι,1+2ι,2+ι,7+0ι,11+0ι,2+3ι,3+2ι]`. What is your opinion?

Math/NumberTheory/GaussianIntegers.hs
```diff
@@ -0,0 +1,233 @@
...
+-- check the precondition.
+findPrime' :: Integer -> GaussianInteger
+findPrime' p =
+    let (Just c) = Moduli.sqrtModP (-1) p
+        k  = Powers.integerSquareRoot p
+        bs = [1 .. k]
+        as = map (\b' -> (b' * c) `mod` p) bs
+        a  = head [ a' | a' <- as, a' <= k]
+        b  = head [ b' | b' <- bs, b' * b' + a * a == p]
+    in a :+ b
```
Bodigrim: IMHO it is more idiomatic to zip `bs` and `as` together, when we need to find correspondent elements. E. g., `abs = map (\b' -> (b' * c `mod` p, b') bs`. 

I am not sure that `head` is a right choice here. If a user supply unsupported prime (of form 4k+3), he will get confusing `Exception: Prelude.head: empty list`. It would be great to show custom error message with a relevant hint.

Math/NumberTheory/GaussianIntegers.hs
```diff
@@ -0,0 +1,233 @@
...
+         , (a', b') <- if p `mod` 4 == 3
+             then [ (p, 0) ]
+             else [ (a, b)
+                  | let radius = Powers.integerSquareRoot p
+                  , a <- [0 .. radius]
+                  , let d = p - a * a
+                  , Powers.isSquare d
+                  , let b = Powers.integerSquareRoot d
+                  ]
+         ]
```
Bodigrim: Can `findPrime` be reused here?

Bodigrim:
Don't worry, rebase is OK.
I have added the final comments, we are almost ready to merge :)

cfredric:
The only comment I disagree with is about the usage of `head` in `findPrime'`. It's documented that `findPrime'` doesn't check the precondition (p is prime and not congruent to 3, mod 4), for use in places like `factorise`  where the precondition has already been checked. `findPrime` does check the precondition, and gives a more helpful error message.

I updated the `Show` instance so that it produces the following (I chose some illustrative cases):
    `[-2-2*ι,-2-ι,-2,-2+ι,-2+2*ι,-1-2*ι,-1-ι,-1,-1+ι,-1+2*ι,-2*ι,-ι,0*ι,ι,2*ι,1-2*ι,1-ι,1,1+ι,1+2*ι,2-2*ι,2-ι,2,2+ι,2+2*ι]`
Let me know what you think.

cartazio:
i've been really enjoying following this thread, and hopefully I'll be able to help do some review tomororw too!

@Bodigrim do you feel comfortable including this in our WIP bug fix extravaganza release? (assuming we all are on board with it?)

Bodigrim:
@cartazio I think it is OK. We planned the next release primarily for bug fix purposes, but it is nice to have some new, covered with tests, high-level code too. 

@cfredric I am extremely happy with your contribution, thank you very much for your patience and your efforts. If no one minds, I am planning to perform a merge in two weeks.

cfredric:
Great! Thanks very much for helping get it ready!

cfredric:
One final question - do you think it's worth defining 'i = 0 :+ 1' as well, in case people don't want to use 'ι'?

Bodigrim:
i is too popular variable name. I do not think it is good to occupy it by library, clashes will be overwhelming.

PR #24, opened 2016-03-16
Improve .travis.yml and prepare for GHC 8.0

Bodigrim:
1. I improved .travis.yml, starting from http://docs.haskellstack.org/en/stable/GUIDE/#travis-with-caching. The main advantage is using cache, which potentially makes builds much faster.

2. GHC 8.0 is landing soon. I have bumped up package dependencies, fixed new warnings `Redundant constraint...`, added new build to Travis. 

Since this is a technical update and no real stuff is affected, I am going to merge it in a few days, if no one minds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,95 +1,94 @@
-# NB: don't set `language: haskell` here
+# Inspired by http://docs.haskellstack.org/en/stable/GUIDE/#travis-with-caching
 
-# See also https://github.com/hvr/multi-ghc-travis for more information
+# Use new container infrastructure to enable caching
+sudo: false
 
-# The following lines enable several GHC versions and/or HP versions
-# to be tested; often it's enough to test only against the last
-# release of a major GHC version. Setting HPVER implictly sets
-# GHCVER. Omit lines with versions you don't need/want testing for.
-env:
- - CABALVER=1.18 GHCVER=7.6.3
- - CABALVER=1.18 GHCVER=7.8.3
- - CABALVER=1.22 GHCVER=7.10.3
- - CABALVER=head GHCVER=head
+# Choose a lightweight base image; we provide our own build tools.
+language: c
+
+# Caching so the next build will be fast too.
+cache:
+  directories:
+  - $HOME/.ghc
+  - $HOME/.cabal
+  - $HOME/.stack
+
+# The different configurations we want to test. We have BUILD=cabal which uses
+# cabal-install, and BUILD=stack which uses Stack. More documentation on each
+# of those below.
+#
+# We set the compiler values here to tell Travis to use a different
+# cache file per set of arguments.
+#
+# If you need to have different apt packages for each combination in the
+# matrix, you can use a line such as:
+#     addons: {apt: {packages: [libfcgi-dev,libgmp-dev]}}
 matrix:
+  include:
+  # We grab the appropriate GHC and cabal-install versions from hvr's PPA. See:
+  # https://github.com/hvr/multi-ghc-travis
+  - env: BUILD=cabal GHCVER=7.6.3 CABALVER=1.16
+    compiler: ": #GHC 7.6.3"
+    addons: {apt: {packages: [cabal-install-1.16,ghc-7.6.3], sources: [hvr-ghc]}}
+  - env: BUILD=cabal GHCVER=7.8.4 CABALVER=1.18
+    compiler: ": #GHC 7.8.4"
+    addons: {apt: {packages: [cabal-install-1.18,ghc-7.8.4], sources: [hvr-ghc]}}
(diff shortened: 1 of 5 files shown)


cartazio:
shouldn't our build matrix do non-stack AND stack builds? (testing only one and not the other seems like it'll miss something)

cartazio:
That is, vanilla cabal  should be testing too, 
i generally try to hew to the example in the multi ghc hvr travis file https://github.com/hvr/multi-ghc-travis/blob/master/.travis.yml and the associated recs in the readme https://github.com/hvr/multi-ghc-travis/blob/master/README.md

though I DO also think the lens projects travis file https://github.com/ekmett/lens/blob/master/.travis.yml is a good role model generally,

why aren't we simply mirroring them?

cartazio:
hrm, looking at how you adapted the travis files, I think maybe a better reference would be the hvr/lens examples, unless i'm misreading the intent :0

.travis.yml
```diff
@@ -1,95 +1,97 @@
...
+
+# Choose a lightweight base image; we provide our own build tools.
+language: c
+
+# Caching so the next build will be fast too.
+cache:
+  directories:
+  - $HOME/.ghc
+  - $HOME/.cabal
+  - $HOME/.stack
```
cartazio: we aren't invoking stack in the build matrix, we can remove this

.travis.yml
```diff
@@ -1,95 +1,97 @@
...
- - export PATH=$HOME/.cabal/bin:/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$PATH
+# Using compiler above sets CC to an invalid value, so unset it
+- unset CC
+
+# We want to always allow newer versions of packages when building on GHC HEAD
+- CABALARGS=""
+- if [ "x$GHCVER" = "xhead" ]; then CABALARGS=--allow-newer; fi
+
+# Download and unpack the stack executable
+- export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$HOME/.local/bin:$PATH
```
cartazio: dead code in build matrix?

.travis.yml
```diff
@@ -1,95 +1,97 @@
...
 
 install:
- - cabal --version
- - echo "$(ghc --version) [$(ghc --print-project-git-commit-id 2> /dev/null || echo '?')]"
- - travis_retry cabal update
- - cabal install --only-dependencies --enable-tests --enable-benchmarks
+- echo "$(ghc --version) [$(ghc --print-project-git-commit-id 2> /dev/null || echo '?')]"
+- if [ -f configure.ac ]; then autoreconf -i; fi
+- |
+  case "$BUILD" in
```
cartazio: see above, stack case is dead code

.travis.yml
```diff
@@ -1,95 +1,97 @@
...
-      cabal install --force-reinstalls "$SRC_TGZ";
-   else
-      echo "expected '$SRC_TGZ' not found";
-      exit 1;
-   fi
-
-# EOF
+- |
+  case "$BUILD" in
+    stack)
```
cartazio: deadcode :)

Bodigrim:
I hesitate whether `arithmoi` gains any profit from both cabal and stack builds (except of heating up Travis clouds). My initial (and rather spontaneous) decision was, as you have spotted from the code, to skip stack builds, but I do not have any real arguments. What is your opinion?

cartazio:
Once the code is on firmer footing, if we re add it back to stack it'll get
plenty of testing by stackage by the stackage maintainers.  I hesitate to
agree to inclusion in stackage LTS, unless we feel confident (via eg all
the awesome new testing and applicable code coverage info we should track
that cabal/hpc can provide us) that it's in a more rock solid place.

Either way, Travis rate limits the build queues for open source projects ,
so unless it allows us to test an env we can't otherwise run ci for...
Stuff.

You can run the .hs file in the hvr repo on our cabal file to generate a
pretty spiffy Travis file.  Then we can eg add testing ghc master to the
build matrix by hand pretty easily.  Or I'm happy to do that run and paste
a straw man on this thread

Bodigrim:
Good idea to run `make_travis_yml.hs` from `hvr/multi-ghc-travis`. I did not know about it. Done.

cartazio:
AWESOME.

looks good to me. there may be ways we can augment or customize it over time, but this is a nice start,  -ff merge at will! :)

cartazio:
@Bodigrim  one last style idea: 
add the   compiler: ": #GHC HEAD" style thingy so the build matrix lists it by name 
as in https://travis-ci.org/cartazio/arithmoi/builds/117130867

Bodigrim:
Nice catch. Fixed.

PR #30, opened 2016-06-15
New arithmetic functions

Bodigrim:
This branch aims to refactor implementation of number-theoretic functions on the modern basis. 

The released `arithmoi` provides a few arithmetic functions, including the divisors function `tau`, the sum of divisors function `sigma`, `totient`, `moebius` and `carmichael`. All of them have monomorphic type `Integer -> Integer`. There are several areas to improve:

1. One may need arithmetic functions, working on other domains. This includes both `Int`, `Word`, `Natural` arguments (where one may expect better performance than in general case), and other unique factorisation domains, such as `GaussianInteger` (and many others in future). We should have a way to define arithmetic function polymorphically.

2. There is no helpers to define new arithmetic functions. Every time one have to write everything from the scratch. Simple wrappers for common classes (multiplicative, additive, etc.) would simplify a deal. 

3. Current functions lacks composability. For example, if I want to compute sum of `totient` and `moebius` on the same argument, the argument will be factored twice, which is expensive. The type of arithmetic function should be `Applicative`. 

The proposed branch addresses the issues above. It includes new `UniqueFactorisation` type class to collect all unique factorisation domains under single interface. It also includes a special type `ArithmeticFunction` with a pack of useful instances and helpers. Both changes introduces additional abstraction layers, but I have managed to minimise their impact on performance. 

My intention was to avoid low-level stuff (unboxed types and tuples) and thoughtless pragmas INLINE / SPECIALISE. `unsafeCoerce` is the only trick, which is used widely to convert between small positive integers of different types. It is ugly, but quite safe and has significant effect on performance. 

Here are benchmarks:

```
benchmarking divisors/old
time                 456.4 ms   (409.3 ms .. 500.3 ms)
benchmarking divisors/new
time                 525.9 ms   (NaN s .. 541.0 ms)
benchmarking divisors/newint
time                 305.9 ms   (301.4 ms .. 308.1 ms)

benchmarking totient/old
time                 185.7 ms   (174.0 ms .. 197.6 ms)
benchmarking totient/new
time                 192.3 ms   (181.4 ms .. 203.8 ms)

benchmarking carmichael/old
time                 200.1 ms   (193.3 ms .. 204.9 ms)
benchmarking carmichael/new
time                 210.2 ms   (203.0 ms .. 216.3 ms)

benchmarking moebius/old
time                 129.7 ms   (121.9 ms .. 145.6 ms)
benchmarking moebius/new
time                 130.3 ms   (125.6 ms .. 134.9 ms)

benchmarking tau/old
time                 183.7 ms   (176.4 ms .. 197.6 ms)
benchmarking tau/new
time                 181.6 ms   (167.0 ms .. 195.8 ms)

benchmarking sigma 1/old
time                 188.2 ms   (180.1 ms .. 195.8 ms)
benchmarking sigma 1/new
time                 186.8 ms   (184.5 ms .. 190.4 ms)

benchmarking sigma 2/old
time                 245.1 ms   (235.7 ms .. 265.8 ms)
benchmarking sigma 2/new
time                 213.3 ms   (197.3 ms .. 226.3 ms)
```


cartazio:
Sweet.  I'll try to take a look asap. Let's find some time to also video
chat synch up soon :)

grandpascorpion:
Hi Andrew,

Very cool.

I noticed one thing recently, that the definition for Integer** is:

data Integer  = S#                !Int#

                -- ^ iff value in @[minBound::'Int', maxBound::'Int']@ range

              | Jp# {-# UNPACK #-} !BigNat

                -- ^ iff value in @]maxBound::'Int', +inf[@ range

              | Jn# {-# UNPACK #-} !BigNat

                -- ^ iff value in @]-inf, minBound::'Int'[@ range

Would it make sense to have any logic to u"nwrap" Integer for performance
gains?

Best, Fred

** There are two GHC.Integer.Type modules defined.(I assume
libraries/integer-gmp/src/GHC/Integer/Type.hs is what's used now as opposed
to libraries/integer-simple/GHC/Integer/Type.hs)

cartazio:
We are aware of how integer Is implemented. :)

The glib but honest answer is: try out the idea you have and also bench
mark it

One issue with what I think you're proposing is that the overflow semantics
for Int... Exist.  But integer / big Nat can't overflow.  So any Int
specialized case needs to keep all the overflow to big nums unless you have
a specialized optimizer that can reason about that.

Amusingly a lot of the bugs @bodigrim fixed in the most recent release were
un checked overflow cases.  I think it's a premature optimization to
introduce that same pattern back without benchmarks that justify it and
explanations of why it isn't buggy / or good coverage ;)

grandpascorpion:
Thanks for the explanation.  Just a thought.
I definitely wasn't implying sticking something just because it "felt good" :)

Andrew and I had a few discussions earlier about the curve testing that didn't end up bearing fruit.

cartazio:
Sorry if I sounded cross.
I guess I prefer having these sorts of exploratory idea storms on their own
ticket / email thread / irc chat or the like.

What might be cool is for you make a ticket that you use log / think out /
track experiment results on as a sort of journal. Or one per idea.  A good
grounding rule of thumb (that I fail to adhere to ) is to make sure you
have one small worked out example exhibiting what you suggest.

Now I hope you don't mind, but let's leave this pr thread for code review :)
(I may delete these comments to clean up the pr thread , hope you don't
mind )

grandpascorpion:
No worries.  You're right

Math/NumberTheory/ArithmeticFunctions/Class.hs
```diff
@@ -0,0 +1,102 @@
...
+-- | A typical arithmetic function operates on the canonical factorisation of
+-- a number into prime's powers and consists of two rules. The first one
+-- determines the values of the function on the powers of primes. The second
+-- one determines how to combine these values into final result.
+--
+-- In the following definition the first argument is the function on prime's
+-- powers, the monoid instance determines a rule of combination (typically
+-- 'Product' or 'Sum'), and the second argument is convenient for unwrapping
+-- (typically, 'getProduct' or 'getSum').
+data ArithmeticFunction n a where
```
cartazio: this is interesting!

Math/NumberTheory/ArithmeticFunctions/Standard.hs
```diff
@@ -0,0 +1,300 @@
...
+#endif
+
+#if MIN_VERSION_base(4,7,0)
+import Data.Coerce
+#else
+coerce :: a -> b
+coerce = unsafeCoerce
+#endif
+
+-- There are special rewrite rules for (^ Int), but not for (^ Word).
```
cartazio: fromIntegral has a good specialization rule for this conversion, I think, as does Data.Bits, perhaps one of those would be what you want?

Bodigrim:
Hmm. Surprisingly I cannot reproduce the performance boost from unsafe coercions between integer types. So I have cleaned `unsafeCoerce` away.

Bodigrim:
Something went wrong on Travis for GHC 7.6: aeson 0.11.2 cannot be built. Very strange: there were no changes in dependencies since previous (successful) Travis build.

cartazio:
Huh.  I'm at some family stuff at the moment  but I'll take a look later
today.

Yeah, unsafe coerce shouldn't really be accomplishing anything here.  Also
the coerce / coercible builtin for recent ghcs should cover most use
cases.  And for standard Int types there should be all sorts of
specializing rules via fromIntegral and friends. I think.

Bodigrim:
Fixed. GHC 7.6 is a real headache. 

Any more comments?

cartazio:
I'll have a look see later tonight or tomorrow.  This week I've been at the
LICS conference a bunch of days / generally catching up on backlog and rest
and rest and rest.  Rest maters.

I do like that you made that abstraction a record instead of a class.  If
only because type class style approach gets into a rabbit hole.

Hrmm ... Do we want to support 7.6 - 8.0 and master for the next release?

Bodigrim:
IMHO we might drop GHC 7.6 support in the future release. I am totally for it: support 7.8, 7.10 and 8.0 should be enough.

Bodigrim:
If there is no objection, I am going to merge this pull request by the end of the week.

.travis.yml
```diff
@@ -76,7 +76,7 @@ install:
 # any command which exits with a non-zero exit code causes the build to fail.
 script:
  - if [ -f configure.ac ]; then autoreconf -i; fi
- - cabal configure --enable-tests --enable-benchmarks -v2 --ghc-options="-Werror" -f check-bounds # -v2 provides useful information for debugging
+ - cabal configure --enable-tests --${BENCH}-benchmarks --constraint="semigroups +bytestring" -v2 --ghc-options="-Werror" -f check-bounds # -v2 provides useful information for debugging
```
cartazio: why do we have constraints flags here?

Bodigrim: Just a debris of experiments around GHC 7.6. Cleaned.

PR #32, opened 2016-08-02
Test and benchmark karatsuba

Bodigrim:
Tests and benchmarks for PR #6 

If no one minds, I am planning to merge it in a week.


cartazio:
Cool! Were there any intersting findings from the benchmarking?

I'll be at HacBoston this weekend and I'll give things at least a skim

Bodigrim:
Benchmarks showed that Karatsuba sqrt is faster than Heron starting from ~2300 bits. Previously the threshold was set to 4096 bits.

PR #33, opened 2016-08-10
Support QuickCheck-2.9

phadej:


phadej:
Now it fails not due me:
```
          quickcheck Large Int:           FAIL
            *** Failed! Exception: 'Math.NumberTheory.Moduli.jacobi: negative denominator' (after 82 tests and 29 shrinks): 
            Large {getLarge = 0}
            Large {getLarge = 4611686018427387904}
```

phadej:
Probably due in `QuickCheck-2.9`
```
* Fixed bugs:
		* arbitrarySizedBoundedIntegral wasn't generating huge integers
```

phadej:
... and other test jobs stall as too big numbers get generated...

phadej:
So I'm not sure how to proceed. We could add a newtype to generate "not too big" numbers, which shouldn't overflow, then we could document those ranges?

Bodigrim:
Thank you. I'll take a look on the broken test, but it won't quick - I am a bit busy this week.

Bodigrim:
1. A bunch of errors are due to integer overflow in `Math.NumberTheory.ModuliTests.toOdd`. Now, when QuickCheck returns decently large numbers, `2*x+1` appears to be negative. The proper solution should introduce a newtype `Odd` with quickcheck/smallcheck instances and some mumbo-jumbo to `Compose` it with other functors (like `NonNegative`).

2. We also use a special newtype wrapper `Huge` to generate really HUGE `Integer`. However, I suppose that with new QuickCheck we can fallback to `Large Integer`, which is hopefully large enough to trigger all sorts of interesting stuff. And `Huge` became just too huge.

3. I am not sure what went wrong with tests for `powerMod`, but I suggest that it is integer overflow again.

I hope to get time for fixes next week, but if anybody wants to help - you are really very welcome.

phadej:
I don't think `Arbitrary Integer` is changed, only bounded Integral types.

Bodigrim:
As discussed above I have created #37.

Bodigrim:
I've fixed issues with tests #37, so that we can return to your PR. 

Are you sure that bumping lower bound of QuickCheck is reasonable? Even nightly Stackage does not have QuickCheck 2.9 yet. I'd rather keep the lower bound at 2.8, bumping up only the upper one.

phadej:
Stackage will eventually catch up, for now they (or we) could disable tests for packages with "positively incompatible" lower-bound

Bodigrim:
Is it very difficult to support both QuickCheck 2.8 and 2.9 for a while?

phadej:
Than you'll need to keep instance around, and have more complicated `CPP`: define instance only if old-base and old-QuickCheck are used. 

IMHO as it's a test dependency, it's not worth the burden.

Bodigrim:
I'd prefer more complicated CPP in this case. Currently HEAD of arithmoi can be build (including tests and benchmarks) by stack using current LTS without a question, which is very convenient, especially for newbie contributors from academia.

phadej:
I'll amend the PR asap

phadej:
I relaxed also some bounds. Builds out of the box with *LTS-2.22* too (the last for GHC-7.8.4)

Bodigrim:
Wow, cool, many thanks.

PR #35, opened 2016-08-16
Bugfix and add quickcheck test for `sieveFactor` trial division

### LINKED ISSUE #7: sieveFactor (factorSieve 2048) (29*73) == [(2117,1)]

ExpHP:
After much effort trying to decipher the trial division loop in `sieveFactor`, the only conclusion I could come to is that it cannot possibly work.  And it does not:

```haskell
import Math.NumberTheory.Primes.Factorisation
main = print $ sieveFactor (factorSieve 15) 75 -- prints [(75,1)]
```

As far as I can surmise, `toPrim` simply does not have anything to do with `FactorSieve` (it is for something that uses a much higher-order wheel optimization).

This commit also adds a quickcheck/smallcheck style test on a fairly small sieve (size 25) in a hope to exercise the trial division and curve factorization paths beyond just singular cases.

---
This raises the question, how did #17 appear to fix #7 if it was still using the wrong indices?

I was curious about this and also wanted to make sure that my "fix" wasn't just masking the issue by e.g. causing it to begin curve factorization too early.  After [debugging it with some traceShows](https://gist.github.com/ExpHP/6d8acbbd070ee8e67c71ebe0d87c8bc5), it appears that the 2048 case was merely fixed "by chance", as the majority of small odd numbers are prime. (so there is a high chance of `pix` fortituously being zero)

### DIFF
--- a/Math/NumberTheory/Primes/Sieve/Misc.hs
+++ b/Math/NumberTheory/Primes/Sieve/Misc.hs
@@ -56,7 +56,7 @@ FactorSieve:
    To relate an odd number x to its index i:
 
        i = (x `div` 2) - 1
-       x = i*2 + 3
+       x = i * 2 + 3
 
 TotientSieve, CarmichaelSieve:
 
@@ -170,7 +170,7 @@ sieveFactor (FS bnd sve) = check
                                                j | j <= bound -> intLoop (fromIntegral (j `shiftR` 1) - 1)
                                                  | otherwise -> tdLoop j (integerSquareRoot' j) (ix+1)
           where
-            p = fromIntegral $ 2*ix + 3
+            p = fromIntegral $ 2 * ix + 3
             pix = unsafeAt sve ix
     curve n = stdGenFactorisation (Just (bound*(bound+2))) (mkStdGen $ fromIntegral n `xor` 0xdecaf00d) Nothing n
 
--- a/Math/NumberTheory/Primes/Sieve/Misc.hs
+++ b/Math/NumberTheory/Primes/Sieve/Misc.hs
@@ -149,8 +149,8 @@ sieveFactor (FS bnd sve) = check
                                                j | j <= bound -> intLoop (fromIntegral (j `shiftR` 1) - 1)
                                                  | otherwise -> tdLoop j (integerSquareRoot' j) (ix+1)
           where
-            p = toPrim ix
-            pix = unsafeAt sve $ fromIntegral p
+            p = fromIntegral $ 2*ix + 3
+            pix = unsafeAt sve ix
     curve n = stdGenFactorisation (Just (bound*(bound+2))) (mkStdGen $ fromIntegral n `xor` 0xdecaf00d) Nothing n
 
 -- | @'totientSieve' n@ creates a store of the totients of the numbers not exceeding @n@.
--- a/Math/NumberTheory/Primes/Sieve/Misc.hs
+++ b/Math/NumberTheory/Primes/Sieve/Misc.hs
@@ -47,6 +47,27 @@ import Math.NumberTheory.Primes.Factorisation.Utils
 import Math.NumberTheory.Unsafe
 import Math.NumberTheory.Utils
 
+{-
+IMPORTANT NOTICE: Not all sieves use the same layout!
+
+FactorSieve:
+
+   To remain as efficient as possible, FactorSieve omits only even numbers.
+   To relate an odd number x to its index i:
+
+       i = (x `div` 2) - 1
+       x = i*2 + 3
+
+TotientSieve, CarmichaelSieve:
+
+   These sieves use a (2,3,5) wheel optimization, sacrificing performance to save
+   more memory. The only indices stored are those coprime to 2, 3, and 5.
+   To relate such an integer x to its index i:
+
+       i = toIdx x
+       x = toPrim i
+-}
+
 -- | A compact store of smallest prime factors.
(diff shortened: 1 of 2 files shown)


Bodigrim:
Wow. I deeply appreciate your contribution. 

Are `FactorSieve` and `TotientSieve / CarmichaelSieve` indexed differently? AFAIU the latter ones are wheel-optimised with `toIdx / toPrim` to convert numbers to indices and back, and the first one use `\x -> 2*x+3` and `\x -> x div 2 - 1`. Am I right? If yes, could you please note this difference near the definitions of `*Sieve` data types?

cartazio:
This also makes me notice a) there's a very suspicious adhoc seed initialization for randomness going on and b) we should get random v2 in shape because the v1 rng stuff has terrible quality.

ExpHP:
I agree it does seem to be that `toIdx` and `toPrim` is meant for the other two sieves.

What has my attention now is [this little bit](https://github.com/cartazio/arithmoi/blame/master/Math/NumberTheory/Primes/Sieve/Misc.hs#L281-L309); it is very undocumented and my only guess from the name is that it is another "smallest prime factor sieve" for use by the other sieves.  Curious... (thankfully Github makes it not too difficult to look into the history behind things like this)

Note: The asymmetry between `\x -> 2*x+3` and `\x -> x div 2 - 1` somehow deeply bothers me; while it does work for odd integers, I cannot possibly imagine why it would possibly be written that way (other than to deliberately confuse the reader!)

Bodigrim:
Feel free to rewrite `x div 2 - 1` as `(x - 3) div 2` :) It is one of many pseudo-micro-optimisations, scattered here and there across `arithmoi`. BTW it would be nice if you get time to write basic benchmarks for these sieves.

ExpHP:
@Bodigrim:

> If yes, could you please note this difference near the definitions of *Sieve data types?

I just added a comment which attempts to explain the differences.  I tried to squeeze in both the differences in usage and the apparent reasons behind them (to my best understanding), hopefully without turning it into a wall of text that nobody will read.  Is this good?

Bodigrim:
Cool, many thanks. Could you please surround `*` with spaces in `2*x` and `x*2`?

Is it possible to replace `factorSieve` with `spfSieve` completely? It would make our code more consistent.

cartazio:
Yeah, we general are in favor of removing any pseudo bench marks 

@Bodigrim I think our next release should be gated on auditing how we use/need randomness and random v2 getting done :)

cartazio:
Pseudo micro optimizations. The current code base has tooooooo many of them to the point it's almost a liability

ExpHP:
> Is it possible to replace factorSieve with spfSieve completely? It would make our code more consistent.

It would be more consistent, but at the same time I have concerns:

**Some refactoring should probably come first.**  Without decent abstractions in place, higher order wheels are harder to use correctly, because more factors are omitted. [sieveTotient](https://github.com/ExpHP/arithmoi/blob/a28fd1783ef00c87f19a800c82fdd1721da5b1fc/Math/NumberTheory/Primes/Sieve/Misc.hs#L203-L211) and [sieveCarmichael](https://github.com/ExpHP/arithmoi/blob/a28fd1783ef00c87f19a800c82fdd1721da5b1fc/Math/NumberTheory/Primes/Sieve/Misc.hs#L258-L278) currently both "reinvent the wheel" (pardon the pun) to strip off factors of 2, 3, and 5, for instance.

(Of course, from the looks of things, a refactoring is probably well-deserved; but I think I'm too afraid to touch most of the code!)

**Performance impact?**  It appears that FactorSieve has been like this for about 4 years; db3dcf808106de71c3a5b05 is the commit that introduced the change.  I definitely agree with the sentiment that some benchmarks would be nice.

cartazio:
i think @Bodigrim  and I both welcome any well motivated breaking changes that end with us having better code, apis and performance (AND correctness). Arithmoi is lacking on any of those  historically .


If theres a better design api, lets not fret about consistency. all we have atm is a pile of algorithms that used too much ` {-# LANGUAGE  MagicHash #-}` one night :)

cartazio:
Zooming out, if @Bodigrim is comfortable with this bug fix, i'm thumbs up (dont have any deep number algorithm thinking bandwidth at the moment)

for evolving stuff, lets kick off a sieve ticket / thread of its own :)

Bodigrim:
@ExpHP well done!

I back everything said by @cartazio here :) Sorry, I am short of time to take a meaningful participation in discussion.

PR #37, opened 2016-08-18
QuickCheck 2.9 fixes

Bodigrim:
QuickCheck 2.9 fixed bug in `arbitrarySizedBoundedIntegral`, which generates really huge numbers now. This fix caused several test failures in `arithmoi` due to integer overflow. See discussion in #33.

The PR fixes our test-suite to be compatible with new version of QuickCheck. 

Since there is no changes in library code, I am keen to merge it ASAP, if no-one minds.

### DIFF
--- a/test-suite/Math/NumberTheory/ModuliTests.hs
+++ b/test-suite/Math/NumberTheory/ModuliTests.hs
@@ -8,8 +8,8 @@
 -- Tests for Math.NumberTheory.Moduli
 --
 
-{-# LANGUAGE CPP          #-}
-{-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE CPP             #-}
+{-# LANGUAGE ViewPatterns    #-}
 
 {-# OPTIONS_GHC -fno-warn-type-defaults #-}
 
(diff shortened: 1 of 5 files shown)


test-suite/Math/NumberTheory/TestUtils/Compose.hs
```diff
@@ -0,0 +1,63 @@
...
+import Control.Applicative
+#endif
+import Control.Arrow
+import Data.Functor.Classes
+import Data.Functor.Compose
+
+import Test.Tasty.QuickCheck (Arbitrary(..))
+import Test.SmallCheck.Series (Serial(..))
+
+instance Num (f (g a)) => Num (Compose f g a) where
```
cartazio: The moment we have orphans for stuff like this I feel like we should just copy compose into our side and have it be a fresh definition.  Else we will get maintenance sadness at some point.

test-suite/Math/NumberTheory/ModuliTests.hs
```diff
@@ -23,51 +23,51 @@ import Control.Arrow
...
-toOdd n = n * 2 + 1
-
 unwrapPP :: (Prime, Power Int) -> (Integer, Int)
 unwrapPP (Prime p, Power e) = (p, e)
 
 -- | Check that 'jacobi' matches 'jacobi''.
-jacobiProperty1 :: (Integral a, Bits a) => AnySign a -> NonNegative a -> Bool
-jacobiProperty1 (AnySign a) (NonNegative (toOdd -> n)) = n == 1 && j == 1 || n > 1 && j == j'
+jacobiProperty1 :: (Integral a, Bits a) => AnySign a -> (Compose NonNegative Odd) a -> Bool
+jacobiProperty1 (AnySign a) (Compose (NonNegative (Odd n))) = n == 1 && j == 1 || n > 1 && j == j'
```
cartazio: Would view patterns or pattern synonyms help this code a teeny bit?
Either way cool use of compose aside from the orphanage issue :

test-suite/Math/NumberTheory/TestUtils/Compose.hs
cartazio: That said, I really like this composite functor model of numbers flavor.  Might be fun to explore further sometime. Not sure if that's an arithmoi layer or numbers package or whatever

Bodigrim:
I have managed to derive all required instances automatically. Hope it looks better now.

Pattern synonyms are cool (see b940f1e), but unfortunately GHC 7.6 does not support them. IMHO we should make one more stable release (0.4.3.0? in September?) and stop supporting GHC 7.6 afterwards. It causes too much trouble, I am really tired of all this mess with `#ifdef` here and there.

cartazio:
I'll take a look after my evening.  Yeah I guess a stable bug fix is a good
idea.

Did you keep it an orphan or add the newtype? ;)

Bodigrim:
I kept it orphan; instances obtained via `StandaloneDeriving` should be safe and maintainable.

cartazio:
Safe in that the build will fail if they ever become not orphan? I suppose
that's true.  I guess I prefer builds for everrr style for libraries :)

Bodigrim:
Frankly speaking, I doubt that avoiding several automatically derived orphan instances justify the introduction of a new newtype and copy-pasting everything from `Data.Functor.Compose`. Since GHC 8.0 `Compose` is in `base`, so its set of instances should not change rapidly. In the **worst** case scenario such changes will break the test suite, but not the library, which I consider not too harmful.

cartazio:
Sounds good :). That's totally solid reasoning and ultimately I agree

Always good to talk out these things explicitly :)

On Aug 25, 2016 2:48 PM, "Bodigrim" <notifications@github.com> wrote:

Frankly speaking, I doubt that avoiding several automatically derived
orphan instances justify the introduction of a new newtype and copy-pasting
everything from Data.Functor.Compose. Since GHC 8.0 Compose is in base, so
its set of instances should not change rapidly. In the *worst* case
scenario such changes will break the test suite, but not the library, which
I consider not too harmful.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/cartazio/arithmoi/pull/37#issuecomment-242498108>, or mute
the thread

phadej:
Can we have a release with this patch?

Bodigrim:
Personally I do not mind, but could you please raise an issue? Is there any particular reason to release right now?

phadej:
https://github.com/fpco/stackage/issues/1691

Issue #40, opened 2016-09-09
New release of `arithmoi`


phadej:
Can the current `master` be released to Hackage. It will help with Stackage's https://github.com/fpco/stackage/issues/1691 going forward (we don't want to disable tests!)

Btw, I checked that `master` is supports all recent versions of other dependencies as well.

Bodigrim:
Are there any other upper bounds to bump?

phadej:
@Bodigrim from my pov everything looks good.

Bodigrim:
Cool. I am for it. I can prepare changelog tomorrow. cartazio?

cartazio:
Sorry for the lag.
Sure. Would this be a minor version or bug fix release atm? Or would it
need be a major version?

Bodigrim:
IMHO since we add new functionality, but do not remove old one, it can be a minor release: 0.4.2.0 -> 0.4.3.0. I'll prepare a pull request.

Bodigrim:
@phadej arithmoi 0.4.3 has been released and uploaded onto Hackage. Please let me know if it plays well with Stackage.

Bodigrim:
Closed after successful merge of fpco/stackage#1889

PR #41, opened 2016-09-12
Release 0.4.3

Bodigrim:
As requested in #40

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,3 +1,17 @@
+0.4.3.0:
+    This release supports GHC 7.6, 7.8, 7.10 and 8.0.
+
+    Add Math.NumberTheory.ArithmeticFunctions with brand-new machinery
+    for arithmetic functions: divisors, tau, sigma, totient, jordan,
+    moebius, liouville, smallOmega, bigOmega, carmichael, expMangoldt (#30).
+    Old implementations (exposed via Math.NumberTheory.Primes.Factorisation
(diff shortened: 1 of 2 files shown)


cartazio:
Sounds good to me, I trust your judgement on this stuff :)

cartazio:
Hrm, it looks like we could get the tests green on ghc master if we did some cpp and ghc options shenanigans.  I think suppressing the simplifiable constraint warning may or may not be useful.  I'm not sure. Or maybe it could be made simpler?

Bodigrim:
AFAIR I had tried to use `MonoLocalBinds` as the warning suggests, but it did not help. And IARC simplifying the instance also triggered another troubles. I do not have GHC HEAD at my workstation, so it is cumbersome to test. 

I am not sure what exactly has changed between GHC 8.0 and GHC HEAD, and whether the warning is genuine or only a quirk of unstable version. That means that adding some cpp magic looks to me a bit premature, but if you know what exactly is going on and how to fix it (more or less :) cleanly - I definitely do not mind.

Otherwise, this is a warning, not an error, so it would not prevent `arithmoi` to build in future. And I hope that we will make another release or two before GHC 8.2 lands :)

cartazio:
... This has to do with a different language issue than that.  But yeah not
a blocker.

It has to do with the redundant constraint warnings and some changes
related to that I think.  Though I thought some of that was going to be
removed from wall in 8.2 so etc etc etc

PR #43, opened 2016-11-03
Drop support of GHC 7.6

Bodigrim:
After release of `arithmoi-0.4.3.0` with stability fixes, IMHO it is time to stop supporting GHC 7.6 and `base-4.6`. It is really outdated and the maintenance is extremely tedious: we need to insert CPP macros here and there to make GHC 7.6 happy. The changelog is a solid evidence: ~100 lines of `#if..#endif` hell were required. 

GHC 7.6 also blocks us from using some dependencies. For example, the upcoming branch with Riemann zeta function inevitably requires `exact-pi >= 0.4.1.1`, which requires GHC 7.8 at least.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,6 @@ before_cache:
 
 matrix:
   include:
-    - env: CABALVER=1.16 GHCVER=7.6.3 BENCH=disable
-      compiler: "GHC 7.6.3"
-      addons: {apt: {packages: [cabal-install-1.16,ghc-7.6.3], sources: [hvr-ghc]}}
     - env: CABALVER=1.18 GHCVER=7.8.4 BENCH=enable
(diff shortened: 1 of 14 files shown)


cartazio:
Dooooo it

Kill it all.

Honestly I'd go as far as to support only 7.10.3 or newer.  Alll other ghc
versions in 7.8 and 7.10 major versions have a bug in the register
allocator for any codes working on doubles and floats st the same time, and
I'm sure there's algorithmic accelerator Tricks we could use to better
approximate or get to the right answer that could benefit from float /
double tricks ?  Idk.

But 7.10.3 is the only one that doesn't have that register allocator bug.

Bodigrim:
My vote is to drop ghc 7.8 support once ghc 8.2 will be released.

PR #44, opened 2016-11-04
Riemann zeta function

Bodigrim:
This PR contains an implementation of [Riemann zeta function](https://en.wikipedia.org/wiki/Riemann_zeta_function) on integer arguments. Riemann zeta function is a crucial ingredient of analytical number theory. 

The implementation is polymorphic by `Floating` type, so end user can use data types with arbitrary floating-point precision instead of `Double`. There is also a function for exact analytical values of zeta function on even arguments.

It would be cool to implement the Riemann zeta function of complex argument (not only integers), but this task requires decent numeric libraries. Namely, I have not found any Haskell implementation of [incomplete gamma function](https://en.wikipedia.org/wiki/Incomplete_gamma_function) in complex domain (not saying about arbitrary precision). Maybe next time :)

### DIFF
--- a/Math/NumberTheory/Recurrencies/Linear.hs
+++ b/Math/NumberTheory/Recurrencies/Linear.hs
@@ -28,7 +28,13 @@ import Numeric.Natural
 import Data.Word
 #endif
 
--- | Infinite sequence of factorials, starting with 0!
+-- | Infinite zero-based table of factorials.
+--
+-- > > take 5 factorial
+-- > [1,1,2,6,24]
+--
+-- The time-and-space behaviour of 'factorial' is similar to described in
+-- "Math.NumberTheory.Recurrencies.Bilinear#memory".
 factorial :: (Num a, Enum a) => [a]
 factorial = scanl (*) 1 [1..]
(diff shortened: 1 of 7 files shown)


Math/NumberTheory/Recurrencies/Linear.hs
```diff
@@ -20,6 +21,20 @@ module Math.NumberTheory.Recurrencies.Linear
...
 import Data.Bits
+import Numeric.Natural
+
+#if MIN_VERSION_base(4,8,0)
+#else
+import Data.Word
+#endif
+
+-- | Infinite sequence of factorials, starting with 0!
+factorial :: (Num a, Enum a) => [a]
```
cartazio: I'm not sure if This has the memory behavior you want. Namely the specializations will be shared across usages. But polymorphic one is a function from the type class dictionaries to the lazy sequence.

cartazio: This perhaps doesn't matter.  But might?

Bodigrim: True, thanks for noticing. There was a comment lost in rebasing.

Issue #25, opened 2016-03-18
Conditions on arguments of isStrongFermatPP 


grandpascorpion:
Hi,

isStrongFermatPP  4 9 returns True, meaning base 9 is a "strong liar" for 4
I discovered this accidentally today. I checked the definition and the algorithm seems to only work properly for odd integers > 3.

For n=1: It hung.
For n=2 and  3: It returned False for some bases

I suggest there be a hidden pre-condition function called within 
isStrongFermatPP and millerRabinV (and perhaps the "new" function)
Then,  an underlying function (isStrongFermatPP') would be called if it's a suitable candidate for Miller-Rabin.  Also, isPrime works for negative or positive integers.  I think isStrongFermatPP and related should as well.

-Fred

Bodigrim:
Documentation says:
> isStrongFermatPP n b tests whether n is a strong Fermat probable prime for base b, where n > 2 and 1 < b < n. The conditions on the arguments are not checked.

`isStrongFermatPP 4 9` violates condition `b < n`, so its behaviour is undefined. Same for `isStrongFermatPP` with `n=1` and `n=2`. Further, `isStrongFermatPP 3 b` is `True` for the only valid value of `b=2`.

However, I do not share an opinion of original author of `arithmoi` about unchecked conditions here and there. IMHO it is almost always worth to add guards, at least when unfiltered input data can cause hangs (as `n=1` in your example). If it is possible to lift obscure restrictions on function arguments without significant performance degradation - it should be done. API must incapsulate inner mechanics as much as possible.

> Also, isPrime works for negative or positive integers. I think isStrongFermatPP and related should as well.

This is controversial. Personally I dislike the behaviour of `isPrime` in the first place. Let us leave it as is for now.

cartazio:
Indeed.  Undefined behavior is not math. And as long as a branch is either
heavily biased so branch predictors work well for its not an innner loop,
branches are a OK.

Our plan for the next release is to try and limit changes to accepting new
algorithms that provided they have some reasonable basic testing and
perhaps benchmarks plus shoring up the soundness and testing of preexisting
parts, which Andrew is doing a heroic job of.

Phrased differently: we want to make sure that there's at least one rock
solid release that's easy for folks with old code to adapt to, and ideally
a minor version bump (though it may have to be a major version number).
That said , I could be missing some context in this thread :)

To step back and zoom out a teeny bit, when we talk about perhaps exposing
internal api components, it would be helpful (at least to me), if we can
[...]

Issue #22, opened 2016-03-14
millerRabinV performance enhancement?


grandpascorpion:
millerRabinV : This routine in Math/NumberTheory/Primes/Testing/Probabilistic.hs
seems quite inefficient.  According to the comments in the code, verifying a prime is actually "prime" with this routine would require testing 196 bases (primes through 1200).

In the last few years, there has been a project: https://miller-rabin.appspot.com/
to identify the maximum accurate solution for # of bases: 1 - 7.
A set of 7 bases exists (2, 325, 9375, 28178, 450775, 9780504, 1795265022) 
that can verify the primality of a number  <= 2^64.

Secondly, I wonder if this logic could be used to improve the performance of isPrime for numbers < 2^64.  Not sure what the cutoff would be.  http://www.trnicely.net/misc/bpsw.html

[Edit: Fixed a few typos]

Bodigrim:
Sounds reasonable, thank you for links. Sorry, I am busy at the moment. It will be great if you have spare time to prepare pull request.

grandpascorpion:
Sorry for the noobie question but which branch should I pull from (710+ testing) ?

Bodigrim:
Pull `master` branch. Other branches are outdated and should be pruned in future.

grandpascorpion:
Thanks!

grandpascorpion:
Shoot, I missed something obvious, that one can specify the number of bases to millerRabinV.  I think a new function would make more sense than updating this one.  Otherwise, the "optimal" benefits would result in a clunky implementation. 

I think that logic could still be adapted for isPrime if there are performance benefits.

Bodigrim:
A new function is not a problem, if it is better than old one. Maybe it is worth to run Miller-Rabin over 7 bases above first and then over specified in the `k` parameter number of consequent primes.

I encourage you to change `isPrime` and make benchmarks. It sounds like a very interesting experiment.

grandpascorpion:
Hmm, I think the new function could just accept n as the sole parameter (if
the intention is to check the minimal set of bases).

Maybe we could do something else, like a hybrid function that combined
"new" and millerRabin,
we could ignore the k (# of bases) param if n <= 2^64 or k <8 and n >
2^64.  Otherwise, we could  "supplement" the 7 base list with other small
primes not in the list.

For instance, the optimal 7 base set is:

2,  325,  9375,  28178, 450775, 9780504, 1795265022
[...]

PR #54, opened 2017-04-02
Improve probabilistic prime testing

Bodigrim:
1. Speed up `isPrime` approximately 2x by usage of GMP builtin for trial division. See benchmarks module.
2. Replace `millerRabinV` implementation with a call to GMP builtin. Speeding up `millerRabinV` is the subject of #22.
3. Since `millerRabinV` no longer calls `isStrongFermatPP` inside a loop, it makes no sense to skip precondition checks in `isStrongFermatPP`, which is the subject of #25.
4. Rewrite comments for `isPrime`, `millerRabinV` and `isStrongFermatPP` to reflect changes above.


cartazio:
And we already require integer Gmp for all users right?

Bodigrim:
Yes, long ago :)

Issue #23, opened 2016-03-16
curveFactorisaton : quirk with selecting "parms"


grandpascorpion:
Hi,

In the internal sub-function "fact" in curveFactorisation in Math/NumberTheory/Primes/Factorisation/Montgomery.hs, there's this line:

mult j <$> fact k (if null pfs then digs+4 else digs)

digs = # of digits and is passed to findParms to return B1, B2 and curve count.
Looks a little awkward, I think it should be incrementing it so that it grabs the next tier.  (For instance, jumping from 15 to 20).    

When factoring a number, it looks like you would see this pattern due to the varying differences between tiers:
8,   12, 16, 20, 24(_), 28, 32, 36, 40, 44 (_)  -- num digits checked in findParms
12, 15, 20, 25, 25,    30, 35, 40,  45, 45     -- "tier" found in testParams
- means that that digit-tier would be executed twice which is clearly incorrect.

For a quick fix, we could change the hard-coded 4
to a conditional: (if digs <= 16 then 4 else 5)

This "jump of 4" logic is also present in the certiFactorisation function.

Bodigrim:
I must confess that I do not know `Montgomery.hs` deeply enough to verify your proposal at a glance. IMHO we have to annotate the algorithm with comments, cover it with tests and benchmarks before making any changes in logic. If you came to know the particulars of the algorithm, could you please write comments and make pull request?

From the other side, I am certain that hardcoded table of B1, B2, CC is a wrong practice. As it was discussed in #21, the real solution is to have a formula to compute them for any number of digits and get rid of tabulated tiers.

grandpascorpion:
Hi Andrew,

No worries, I'm not an expert myself.

One key section from
http://www.mersennewiki.org/index.php/Elliptic_Curve_Method sheds some
light:

... As described above it probably sounds a bit of a hit-or-miss affair,
but there are controls built into the programme. By carefully determining
the starting values for the programme it is possible to limit our search to
factors of no more than some size measured in bits (a bit is the smallest
unit of computer memory). After running the requisite number of curves
(detailed on a table on this page) at one particular bit level, we can then
start to look for factors at a higher bit level. Even at the higher bit
level it is still possible to find factors of the smaller size.
 I realize this will sound anecdotal but ... from www.mersenneforum.org, I
know that people would attack a particular problem in groups by curve
count.  Say we had a large composite number, a person might try to run the
recommend number of curves for each digit tier (with the recommended B1, B2
[...]

Issue #21, opened 2016-03-14
Elliptic curve testing (using GMP-ECM's B2 and curve counts)


grandpascorpion:
Hi,

I noticed that GMP-ECM (and other code bases apparently) uses different B2 and curve count values:
http://www.mersennewiki.org/index.php/Elliptic_Curve_Method

Given it's GMP-ECM, I would assume these figures are the best known.

(The B2 and curve counts are the 3rd and 4th members of each tuplet in testParms list in Math/NumberTheory/Primes/Factorisation/Montgomery.hs)   

The arithmoi code has an entry for 12 (digit factors).   The GMP-ECM tables do not.  Is the 12 entry really needed? Oddly, there is "no" optimal number of curves for 15 in the GMP-ECM.  Not sure how to interpret that.

Do you think it's worth changing the arithmoi code?

Bodigrim:
Thank you for your question. 

Currently `arithmoi` uses B1 and curve counts from https://www.alpertron.com.ar/ECM.HTM for 15..50 digits. Alas, I do not know the source of parameters for 55..60 digits.

The entry for 12 digits is not really important, because it is weird to factorise such small numbers via ECM. IMHO for the same reason it makes no harm.

Choice of B2 is not a dogma, usually it is chosen to spend the same magnitude of time in stage 2 as in stage 1. GMP-ECM source code says:
```c
  /* set second stage bound B2: when using polynomial multiplication of
     complexity n^alpha, stage 2 has complexity about B2^(alpha/2), and
     we want stage 2 to take about half of stage 1, thus we choose
     B2 = (c*B1)^(2/alpha). Experimentally, c=1/4 seems to work well.
     For Toom-Cook 3, this gives alpha=log(5)/log(3), and B2 ~ (c*B1)^1.365.
     For Toom-Cook 4, this gives alpha=log(7)/log(4), and B2 ~ (c*B1)^1.424. */
```
This means that the value of `c` is implementation dependent. In order to change current values of B2 with a good reason we need to write benchmarks to determine `c`. Pull requests are welcome :) We use `criterion` to write benchmarks, I'll be glad to help with technical details.

It seems reasonable to choose curve count in the same way as GMP-ECM does (`print_expcurves` in `ecm.c`). Again, pull requests are very welcome.

grandpascorpion:
Thanks for the detail

Bodigrim:
Closed in #65.

PR #56, opened 2017-04-23
Modular arithmetic

Bodigrim:
This branch has started as an attempt to speed up the implementation of factorization on elliptic curve (#21, #23). The algorithm requires many reductions modulo some big integer. Soon enough I realised that writing them manually is very tedious. E. g., one can write

    a * b * c `mod` m,

but if numbers are big enough, it is preferable to perform reductions after each multiplication. So we write

    a * b `mod` m * c `mod` m.

But if `a`, `b` and `c` were not reduced beforehand and are larger than `m`, it is even better to write

    (a `mod` m) * (b `mod` m) `mod` m * (c `mod` m) `mod` m.

When the code is so submerged under the weight of backticks and brackets, only the brave heart would dare to refactor it. The aim of the proposed branch is to make things easier in two ways:

* Reductions are performed behind the scenes in `Num` instance.
* Type system prevents adding numbers, reducted modulo different numbers. This is a common source of errors.

Despite the initial target was factorization algorithm, I have postponed it to a separate branch: [draft](https://github.com/Bodigrim/arithmoi/tree/montgomery2).

---

The solution is to define a type, which keeps its modulo as a type-level natural:

```haskell
newtype Mod (m :: Nat) = Mod Integer
  deriving (Eq, Ord)
```

The `Num` instance is straightforward:

```haskell
instance KnownNat m => Num (Mod m) where
  mx@(Mod x) * Mod y = Mod $ x * y `mod` natVal mx
  mx@(Mod x) + Mod y = Mod $ if xy >= m then xy - m else xy
    where
      xy = x + y
      m = natVal mx
  negate mx@(Mod x) = Mod $ if x == 0 then 0 else natVal mx - x
  abs = id
  signum = const $ Mod 1
  fromInteger x = mx
    where
      mx = Mod $ fromInteger $ x `mod` natVal mx
```

We always keep the value in normalized form between `0` and `m-1`, because it allows to use derived `Eq` and `Ord` instances.

There are also `invertMod :: Mod m -> Maybe (Mod m)` and `powMod :: Mod m -> a -> Mod m` functions, expressed in terms of GMP calls, which is faster than previous manual implementations.

---

The next abstraction is GADT `SomeMod`. Its design was partially inspired by [PARI/GP](https://pari.math.u-bordeaux.fr/doc.html). (In fact, many my design decisions are inspired by PARI/GP, which is as wonderful for number-theory, as it is lacking type discipline and containers.)

The first constructor is a standard approach, hiding type parameter (as in `KnownNat/SomeNat`, `KnownSymbol/SomeSymbol`).

```haskell
data SomeMod where
  SomeMod :: KnownNat m => Mod m -> SomeMod
```

Now we define `modulo` function to construct `Mod`, when modulo is not known at compile time:

```haskell
modulo :: Integer -> Natural -> SomeMod
modulo n m = case someNatVal (toInteger m) of
  Nothing                       -> error "modulo: negative modulo"
  Just (SomeNat (_ :: Proxy t)) -> SomeMod (fromInteger n :: Mod t)
```

It is possible (and actually quite useful) to define arithmetic operations on `SomeMod` by rule

```haskell
(Mod x1 :: Mod m1) `f` (Mod x2 :: Mod m2) = Mod (x1 `f` x2) :: Mod (m1 `gcd` m2).
```

E. g.,

```haskell
2 `modulo` 10 + 4 `modulo` 15 = 1 `modulo` 5
```

But we still cannot define `Num` instance, because it is impossible to implement righteous `fromInteger` function: it takes only one argument, but to construct `SomeMod` both value and modulo are required. This is resolved by adding another constructor `InfMod` to `SomeMod` GADT and some plumbing.

---

Benchmarks show that `isStrongFermatPP` became 2x-6x faster (due to use of GMP built-in for modular power). The performance of `factorise` on large numbers became worse, but its current implementation is slow as hell any way. Its performance is the subject of [montgomery2](https://github.com/Bodigrim/arithmoi/tree/montgomery2) branch.

---

At the final stage of the work on this branch I have found [modular-arithmetic](https://github.com/TikhonJelvis/modular-arithmetic) package, which implements a similar approach to modular arithmetic. (All coincidences of names are accidental.) The main advantages of the proposed PR are:

* Usage of fast GMP calls to invert modular values.
* Implementation of arithmetic on `SomeMod`.
* `Mod` is an instance of `Fractional` and is not an instance of `Integral` (see the issue TikhonJelvis/modular-arithmetic#8 by @CarlEdman)
* Last, but not least: there are tests!


cartazio:
Very cool.

This sounds like it merits a nice design and code review. Please holler my
way if I'm being Too slow about it.

Off the cuff I think you actually want to look at the reflection package by
ed. It allows you to have the shared modulus be selected at runtime but
still be type indexed.

Which might let you drop / cleanup the design I think ?

cartazio:
yeah .. looks like Reflection would be handy here

something like 

`Reifies Nat s => Num (Mod s)` perhaps?
or something like that

Bodigrim:
`reflection` is very handy in more complex cases, when required instances cannot be simply parametrised by type. Here choice of `Num` and `Fractional` instances is done entirely on basis of type-level `Nat`, and it is a built-in kind with handy helpers from `GHC.TypeLits`.

cartazio:
hrmmm... maybe? 
please message/email me a reminder to have a go at doing a fork that ports this PR to that style this weekend or next week so we can compare them hand to hand or i at least try and then give up and agree :)

cartazio:
I'll have a go  this evening / tomorrow at this :)

cartazio:
It's on the top of my todo queue this week :)

Bodigrim:
If there will be no suggestions, I'd like to merge the branch by the end of the month.

cartazio:
Derp yeah. Sorry :)
I'm suspecting you're right.

cartazio:
sounds good to me, i'll still give it a review / go over once some work deadlines clear up :)

PR #58, opened 2017-07-19
Remove a function composition from totientSum by using a better choice for triangle

incertia:
This also better matches the formula described in [mathworld](http://mathworld.wolfram.com/TotientSummatoryFunction.html).
On the other hand, ``totientSum = (`div` 2) . (+1) . generalInversion ((^2) . fromIntegral)`` is another candidate by following the formula from [wikipedia](https://en.wikipedia.org/wiki/Euler%27s_totient_function#Other_formulae), which kills a where clause as well.

The latter *might* offer very minor performance improvements due to shuffling around of arithmetic operations.

I chose to keep using the triangle inversion formula because that was the previously used formula.

### DIFF
--- a/Math/NumberTheory/MoebiusInversion.hs
+++ b/Math/NumberTheory/MoebiusInversion.hs
@@ -23,11 +23,14 @@ import Math.NumberTheory.Unsafe
 
 -- | @totientSum n@ is, for @n > 0@, the sum of @[totient k | k <- [1 .. n]]@,
 --   computed via generalised Moebius inversion.
---   Arguments less than 1 cause an error to be raised.
+--   See <http://mathworld.wolfram.com/TotientSummatoryFunction.html> for the
+--   formula used for @totientSum@.
 totientSum :: Int -> Integer
-totientSum = (+1) . generalInversion (triangle . fromIntegral)
+totientSum n
+  | n < 1 = 0
+  | otherwise = generalInversion (triangle . fromIntegral) n
   where
-    triangle n = (n*(n-1)) `quot` 2
+    triangle k = (k*(k+1)) `quot` 2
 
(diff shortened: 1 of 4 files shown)


cartazio:
pardon my possible confusion, but what does this commit achieve? could you walk me through your reasoning a bit more?

Bodigrim:
Well, it is a small, but still valuable improvement IMO. To make it even more valuable, @incertia, could you please also add case `totientSum n = 0` for non-positive `n` (instead of internal error, which is current behaviour!) and reflect this change in comment and in tests? 

I'd rather keep version with `triangle`, and add a link to the source of formula in comment.

incertia:
all it really does is avoid a composition with (+1) in the definition of totientSum and match the formula in the first search result for "euler totient sum" to avoid any possible confusion if someone examines the source and doesn't recognize the formula used

apart from that, not much.

incertia:
I made the changes @Bodigrim mentioned as well as added two special test cases (totientSum 0 should be 0, and totientSum (negative number) should be 0). Currently the cabal build will output a warning about how the binding for n in triangle shadows the binding in totientSum. If you guys want to merge the changes I can squash the commits and rename n -> k for triangle.

cartazio:
oh ok, i read the diff wrong, i thought you had added the (+1) . :)))))

my bad!

i leave @Bodigrim  to make the call as he sees fit :)

Bodigrim:
@incertia please use [haddock markup ](https://www.haskell.org/haddock/doc/html/ch03s08.html#idm140354810727840) for URLs.

incertia:
I added angle brackets around the URLs and force pushed a squash.

Bodigrim:
Great! Thanks for your contribution.

PR #65, opened 2017-08-16
Factorisation over elliptic curves

Bodigrim:
The aim of this branch is to refactor computational core of factorisation over Montogomery elliptic curve. Here is a changelog:

* Move computations on elliptic curve into separate module. Use more sophisticated types to represent points on a curve, which helps us to avoid mixing points from different curves.
* Write separate tests for arithmetic over elliptic curve.
* #21 Use B1, B2 and CC parameters as per GMP ECM table.
* #23 Fix jumping from tier to tier.
* Rewrite and simplify [Step 1](http://www.mersennewiki.org/index.php/Elliptic_Curve_Method#Step_1) of elliptic factorization. The code was overengineered and prematurely optimized, making it very hard to read and understand. It appeared that all these complications have no measurable impact on performance.
* Reimplement [Step 2](http://www.mersennewiki.org/index.php/Elliptic_Curve_Method#Step_1) in a right way. In fact previous implementation has nothing to do with Step 2, it was actually slightly modified Step 1. This has a great impact on performance, which is witnessed by the following table. For integers, which factors were found on Step 2, timings are drastically lower.

  Benchmark    | v. 0.5.0.1 | branch
  -------------|-----------:|-------:
  factorise50  |     26 ms  |   8 ms
  factorise60  |      6 ms  |   5 ms
  factorise70  |   3053 ms  | 213 ms
  factorise80  |    488 ms  | 169 ms
  factorise90  |     44 ms  |  26 ms
  factorise100 |   4373 ms  |3039 ms

  Nevertheless, there is plenty of room for further improvement. I have done some experiments, but their proper productionalization will take much more time than I have available at the moment. That said, I decided to delay them indefinitely and merge present state of affairs into master.

### DIFF
--- a/Math/NumberTheory/Curves/Montgomery.hs
+++ b/Math/NumberTheory/Curves/Montgomery.hs
@@ -1,0 +1,160 @@
+-- |
+-- Module:      Math.NumberTheory.Curves.Montgomery
+-- Copyright:   (c) 2017 Andrew Lelechenko
+-- Licence:     MIT
+-- Maintainer:  Andrew Lelechenko <andrew.lelechenko@gmail.com>
+-- Stability:   Provisional
+-- Portability: Non-portable (GHC extensions)
+--
+-- Arithmetic on Montgomery elliptic curve.
+--
+
+{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE GADTs               #-}
+{-# LANGUAGE KindSignatures      #-}
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE MagicHash           #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Math.NumberTheory.Curves.Montgomery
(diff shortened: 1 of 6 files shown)


cartazio:
This sounds great.  I'll try to have a look see. 

@ebutleriv might be a good second reader since he was messing around with using some Haskell elliptic curve apis last spring.

EButlerIV:
Oh, nice! Looking at this now

cartazio:
@Bodigrim  question: why do some of the knownnat variables have funny names like "a24"?

:)

cartazio:
Derp never mind. I need to read more closely :)

Bodigrim:
@EButlerIV thanks for looking into this.

EButlerIV:
Np. This looks really good. The new Montgomery curve module is super clean and easy to noodle around with and congrats on the significant factorization performance increase. IMO it's good to merge as-is.

Bodigrim:
Cool. Many thanks for your time.

PR #66, opened 2017-08-31
Use GHC.TypeNats instead of GHC.TypeLits

Bodigrim:
GHC 8.2 features new `GHC.TypeNats` module, providing a total bijection between usual and type-level naturals. Compare it with `GHC.TypeLits` (which is a thin compatibility layer from now on):

```haskell
GHC.TypeLits.someNatVal :: Integer -> Maybe SomeNat
GHC.TypeNats.someNatVal :: Natural -> SomeNat
```

The latter is much more convenient in applications and allows to avoid "impossible" cases. Here is an example from `Math.NumberTheory.Moduli.Class`:

```diff
 modulo :: Integer -> Natural -> SomeMod
-modulo n m = case someNatVal (toInteger m) of
-  Nothing                       -> error "modulo: negative modulo"
-  Just (SomeNat (_ :: Proxy t)) -> SomeMod (fromInteger n :: Mod t)
+modulo n m = case someNatVal m of
+  SomeNat (_ :: Proxy t) -> SomeMod (fromInteger n :: Mod t)
```

The purpose of the branch is to switch to `GHC.TypeNats` for GHC >= 8.2 and provide compatibility layer for older GHCs.

### DIFF
--- a/GHC/TypeNats/Compat.hs
+++ b/GHC/TypeNats/Compat.hs
@@ -1,0 +1,36 @@
+{-# LANGUAGE CPP         #-}
+
+{-# OPTIONS_HADDOCK hide #-}
+
+#if MIN_VERSION_base(4,10,0)
+
+module GHC.TypeNats.Compat
+  ( module GHC.TypeNats
+  ) where
+
+import GHC.TypeNats
+
+#else
+
+module GHC.TypeNats.Compat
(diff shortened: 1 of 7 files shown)


cartazio:
We have type nats in ghc now? Yay

PR #67, opened 2017-09-01
add fsLookup

### LINKED ISSUE #34: Getting the smallest prime factors from a FactorSieve
It does not seem that there is an inexpensive way to look up the contents of a `FactorSieve`, which is unfortunate because a factor sieve can have other useful purposes beyond generating prime factorizations.

Due to memory optimizations I imagine that directly exposing the container is out of the question - but even just a lookup function like `sieveSmallestFactor :: Sieve -> Integer -> Integer` (or `Maybe Integer`) would be immensely useful.

Of course, there are the edge cases of how to handle inputs of `0` or `1`.

ExpHP:
closes #34 

**This commit is broken** (only because Misc isn't exposed), I'm just putting it up to remind myself to come back to it and possibly gather ideas.

I only just now (re)discovered that Primes.Sieve.Misc is not exposed, which means a number of good things I can do:

* I can expose this, `fsBound`, and possibly `fsPrimeTest` from `Factorisation` where they can sit next to `factorSieve`.
* I can also name them `sieveLookup` and etc. for consistency with `sieveFactor`
* There's no need to feel commited to any unusual design choices particular to fsPrimeTest.

### DIFF
--- a/Math/NumberTheory/Primes/Factorisation.hs
+++ b/Math/NumberTheory/Primes/Factorisation.hs
@@ -27,6 +27,7 @@ module Math.NumberTheory.Primes.Factorisation
     , FactorSieve
     , factorSieve
     , sieveFactor
+    , sieveSmallestFactor
       -- *** Trial division
     , trialDivisionTo
       -- ** Partial factorisation
--- a/Math/NumberTheory/Primes/Sieve/Misc.hs
+++ b/Math/NumberTheory/Primes/Sieve/Misc.hs
@@ -21,9 +21,9 @@ module Math.NumberTheory.Primes.Sieve.Misc
       -- ** Smallest prime factors
     , factorSieve
     , sieveFactor
+    , sieveSmallestFactor
     , fsBound
     , fsPrimeTest
-    , fsLookup
       -- ** Totients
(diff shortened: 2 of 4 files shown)


ExpHP:
note: I did push this in a rush.  The reason it's broken is because I did something dumb and changed it (moved it from `Factorisation` where I initially put it to the `Misc` module) right after testing that it worked. (and then when I rediscovered that Misc is private, it opened up a whole new realm of possibilities for better implementations that I want to consider)

ExpHP:
**Todo:**

- [ ] treat edge cases more similarly to `sieveFactor` instead of `fsPrimeTest`
- [ ] expose from `Factorisation`
- [ ] fix test
- [ ] bench against the fastest currently existing alternative to make sure this is at least *worth* it
- [ ] audit docs
  - [ ] incorrect method names
  - [ ] rustisms (backticks, `Option/Some/None`)
  - [ ] `n=0` -> `n == 0`

ExpHP:
[This is not fast enough to be worth it IMO](https://github.com/cartazio/arithmoi/issues/34#issuecomment-326764873)

PR #72, opened 2017-10-01
Lazy factorisation of huge numbers

Bodigrim:
Function `factorise` is a lazy producer of small factors, but not of bigger ones. If trial division is not enough, factorisation over elliptic curve is invoked, which does not return any factor, until it finds them all. This is inconvenient and redundant in many scenarios, when computations can short-circuit as soon as several first factors were found (for instance, Mobius function). 

The proposed patch eliminate this defect. E. g., now it is possible to compute first factor of huge composite without factoring it completely:

```haskell
> head $ factorise' $ 14145130711 * 10000000000000000000000000000000000000121 * 100000000000000000000000000000000000000000000000447
(14145130711,1)
```

The only (and expected) shortcoming is that the list of factors is no longer sorted. But we never claimed it is, it has always been just an implementation detail.

### DIFF
--- a/Math/NumberTheory/GCD.hs
+++ b/Math/NumberTheory/GCD.hs
@@ -260,6 +260,8 @@ cw32 (W32# x#) (W32# y#) = coprimeWord# x# y#
 -- having to merge multiplicities of primes, which occurs more than in one
 -- composite factor.
 --
+-- > > splitIntoCoprimes [(140, 1), (165, 1)]
+-- > [(5,2),(28,1),(33,1)]
 -- > > splitIntoCoprimes [(360, 1), (210, 1)]
 -- > [(2,4),(3,3),(5,2),(7,1)]
 splitIntoCoprimes :: (Integral a, Num b) => [(a, b)] -> [(a, b)]
--- a/Math/NumberTheory/GCD.hs
+++ b/Math/NumberTheory/GCD.hs
@@ -20,11 +20,17 @@
 --
 -- When using this module, always compile with optimisations turned on to
 -- benefit from GHC's primops and the rewrite rules.
-{-# LANGUAGE CPP, BangPatterns, MagicHash #-}
+
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP          #-}
+{-# LANGUAGE LambdaCase   #-}
+{-# LANGUAGE MagicHash    #-}
+
 module Math.NumberTheory.GCD
     ( binaryGCD
     , extendedGCD
     , coprime
+    , splitIntoCoprimes
     ) where
 
 import Data.Bits
@@ -245,3 +251,31 @@ cw16 (W16# x#) (W16# y#) = coprimeWord# x# y#
 
 cw32 :: Word32 -> Word32 -> Bool
 cw32 (W32# x#) (W32# y#) = coprimeWord# x# y#
+
+-- | The input list is assumed to be a factorisation of some number
+-- into a list of powers of (possibly, composite) non-zero factors. The output
+-- list is a factorisation of the same number such that all factors
+-- are coprime. Such transformation is crucial to continue factorisation
+-- (lazily, in parallel or concurrent fashion) without
+-- having to merge multiplicities of primes, which occurs more than in one
+-- composite factor.
+--
+-- > > splitIntoCoprimes [(360, 1), (210, 1)]
+-- > [(2,4),(3,3),(5,2),(7,1)]
+splitIntoCoprimes :: (Integral a, Num b) => [(a, b)] -> [(a, b)]
+splitIntoCoprimes = \case
+  [] -> []
+  ((1,  _) : rest) -> splitIntoCoprimes rest
+  ((x, xm) : rest) -> case popSuchThat (\(r, _) -> gcd x r /= 1) rest of
+    Nothing            -> (x, xm) : splitIntoCoprimes rest
+    Just ((y, ym), zs) -> let g = gcd x y in splitIntoCoprimes
+      ((g, xm + ym) : (x `quot` g, xm) : (y `quot` g, ym) : zs)
+
+popSuchThat :: (a -> Bool) -> [a] -> Maybe (a, [a])
+popSuchThat predicate = go
+  where
+    go = \case
+      [] -> Nothing
+      (x : xs)
+        | predicate x -> Just (x, xs)
+        | otherwise   -> fmap (fmap (x :)) (go xs)
--- a/Math/NumberTheory/Primes/Factorisation/Montgomery.hs
+++ b/Math/NumberTheory/Primes/Factorisation/Montgomery.hs
@@ -51,7 +51,7 @@ module Math.NumberTheory.Primes.Factorisation.Montgomery
 
 import Control.Arrow
 import System.Random
-import Control.Monad.State.Strict
+import Control.Monad.State.Lazy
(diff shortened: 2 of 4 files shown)


Math/NumberTheory/GCD.hs
```diff
@@ -245,3 +251,31 @@ cw16 (W16# x#) (W16# y#) = coprimeWord# x# y#
...
+
+-- | The input list is assumed to be a factorisation of some number
+-- into a list of powers of (possibly, composite) non-zero factors. The output
+-- list is a factorisation of the same number such that all factors
+-- are coprime. Such transformation is crucial to continue factorisation
+-- (lazily, in parallel or concurrent fashion) without
+-- having to merge multiplicities of primes, which occurs more than in one
+-- composite factor.
+--
+-- > > splitIntoCoprimes [(360, 1), (210, 1)]
```
cartazio: took me a moment to think through this, but its nicely written

Math/NumberTheory/Primes/Factorisation/Montgomery.hs
```diff
@@ -152,75 +155,88 @@ stdGenFactorisation primeBound sg digits n
...
 --   unlikely to succeed if @n@ has more than one (really) large prime factor.
-curveFactorisation :: Maybe Integer                 -- ^ Lower bound for composite divisors
-                   -> (Integer -> Bool)             -- ^ A primality test
-                   -> (Integer -> g -> (Integer,g)) -- ^ A PRNG
-                   -> g                             -- ^ Initial PRNG state
-                   -> Maybe Int                     -- ^ Estimated number of digits of the smallest prime factor
-                   -> Integer                       -- ^ The number to factorise
-                   -> [(Integer,Int)]               -- ^ List of prime factors and exponents
+curveFactorisation
+  :: forall g.
```
cartazio: i like that this is parameterized over the RNG alg :)

cartazio: to clarify the comment on this function: the integer input to curveFactorization assumes "small" factors are absent from the factorization of the input, right?

Math/NumberTheory/Primes/Factorisation/Montgomery.hs
```diff
@@ -152,75 +155,88 @@ stdGenFactorisation primeBound sg digits n
...
+                fmap mconcat $ flip mapM cs $
+                  \(x, xm) -> if ptest x
+                              then pure (Factors [(x, xm)] [])
+                              else repFact x b1 b2 (count - 1)
+
+data Factors = Factors
+  { _primeFactors     :: [(Integer, Int)]
+  , _compositeFactors :: [(Integer, Int)]
+  }
+
```
cartazio: the safety/correctness condition for this semigroup seems tricky to enforce outside of internals, eg if 2 is a prime factor and 4 is a composite factor, right?

cartazio: (granted any checking / soundness guarantees would be bad for algorithm complexity)

Math/NumberTheory/Primes/Factorisation/Montgomery.hs
```diff
@@ -152,75 +155,88 @@ stdGenFactorisation primeBound sg digits n
...
+-- 1) for each argument no prime factor divides any composite factor.
+-- 2) numbers, represented by arguments, are coprime.
+instance Semigroup Factors where
+  Factors pfs1 cfs1 <> Factors pfs2 cfs2
+    = Factors (pfs1 <> pfs2) (cfs1 <> cfs2)
+
+instance Monoid Factors where
+  mempty = Factors [] []
+  mappend = (<>)
+
```
cartazio: could we make this code cleaner by considering a Generalized Version of the Factors data type thats some sort of Bifunctor/ProFunctor/SemiGroupoid/Category?

the modify Powers code is suggestive of that in my mind

cartazio: bifunctor would give bimap being the obvious one, just not sure if that improves other bits :)

cartazio: eg 
```
data Factors a b = Factors [(a,Int)] [(b,Int)] 
```
-- because we're assuming exponents x such that abs x \leq 2^31 ?

cartazio: ... should/could the powers be both in Int vs Integer form for generality?

Math/NumberTheory/Primes/Factorisation/Montgomery.hs
Bodigrim: In fact, this `Semigroup` is lawful in any case, it just may be less suitable for factorisation. I have moved comments to the relevant pieces of `curveFactorisation`.

Math/NumberTheory/Primes/Factorisation/Montgomery.hs
Bodigrim: It does not make much sense to make powers anything else then `Int`: even `2 ^ (maxBound :: Int)` is tremendously huge. So I am not sure whether it will make code clearer. 

W. r. t. to `Semigroup` / `Monoid` instances,  AFAIK there is no `DeriveMonoid` language extension to scrap this boilerplate.

Math/NumberTheory/Primes/Factorisation/Montgomery.hs
Bodigrim: I have expanded the comment.

Math/NumberTheory/GCD.hs
Bodigrim: I have added one more (less complicated) example.

Bodigrim:
Thanks for review! I have implemented some of suggestions.

Math/NumberTheory/GCD.hs
cartazio: ok, that def helps

Math/NumberTheory/Primes/Factorisation/Montgomery.hs
cartazio: ok, very nice

what RNG is default STD gen? :)

cfredric:
In the spirit of laziness, the `factorise` in `Math.NumberTheory.GaussianIntegers` could also be improved. The current implementation outputs factors in order of ascending magnitude, but that also isn't guaranteed by the doc comment - it's just an implementation detail.

I'd have to test it out to be sure (since it's been a while since I wrote that code), but I think this would be as simple as removing the `reverse .` in the definition of `factorise` (once `Factorisation.factorise` is modified to be lazy).

cfredric:
Ahh I lied, it's not that simple. It would also require modifying `factorise`'s `helper` to just add factors to the list as it goes, rather than building the list as an accumulator. But I have no doubt that's a trivial change compared to the rest of this PR :)

Math/NumberTheory/Primes/Factorisation/Montgomery.hs
Bodigrim: `System.Random.mkStdGen` with an xored argument of `factorise` as a seed.

Bodigrim:
@cfredric it would be nice if you get time to prepare a patch for `Math.NumberTheory.GaussianIntegers.factorise`.

Bodigrim:
Patch for lazy `factorise` for `GaussianIntegers` is in #76.

PR #76, opened 2017-10-09
Make GaussianIntegers.factorise lazy

cfredric:
This PR makes the changes necessary for GaussianIntegers.factorise to be lazy. It also includes some minor refactoring in GaussianIntegers.factorise and an unexported helper function.


Bodigrim:
Great! Could you please add a test? Kind of https://github.com/cartazio/arithmoi/blob/master/test-suite/Math/NumberTheory/Primes/FactorisationTests.hs#L50

LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2011 Daniel Fischer, 2016-2017 Andrew Lelechenko, Carter Schonwald
+Copyright (c) 2011 Daniel Fischer, 2016-2017 Andrew Lelechenko, Carter Schonwald, Google Inc.
```
Bodigrim: Please bring it into sync with a copyright in https://github.com/cfredric/arithmoi/blob/5698fd8a827c3d123ce6452683310ab4936bcaaf/Math/NumberTheory/GaussianIntegers.hs#L3

cfredric:
Happy to add tests, but I'm not sure how to yet - I'm not sure I see how that laziness test works. It looks like it takes a large number and a (potentially incomplete) list of its factors, (lazily) factorizes the large number, and asserts that the list of provided factors is the same as the front of the list of found factors. But wouldn't that succeed even if `factorise` were strict? Is the idea that the testcase would get killed if it took too long?

(I also pulled the laziness changes into my branch and tried (with `factorise` imported from `Math.NumberTheory.Primes`) computing `head . factorise $ 14145130711 * 10000000000000000000000000000000000000121 * 100000000000000000000000000000000000000000000000447`, but it hasn't yet yielded a value, so it doesn't look lazy to me. Am I missing something?)

Bodigrim:
Yes, `factorise` is not able to handle such huge number in a reasonable time. Works fine in `ghci` for me:

```haskell
$ stack repl --ghci-options -fobject-code
> :set +s
> head . Math.NumberTheory.Primes.factorise $ 14145130711 * 10000000000000000000000000000000000000121 * 100000000000000000000000000000000000000000000000447
(14145130711,1)
(1.30 secs, 1,021,791,680 bytes)
```

cfredric:
I'm stumped on this - I can't get `Math.NumberTheory.Primes.factorise` to be lazy. I've rebuilt `arithmoi`, and tried a variety of ways of importing Math.NumberTheory.Primes and of referring to `factorise`, just to ensure that I'm calling the right version of the function (i.e. not the `factorise` in GaussianIntegers, and not the `factorise` included in the version of `arithmoi` I installed from Hackage), and all of them have failed to produce a value even after a few hours. `git log` shows that I have indeed pulled in the relevant changes, so I'm not sure what else to try.

Bodigrim:
What happens when you run the test suite? Is the correspondent test freezing or failing?

cfredric:
The test suite was passing happily, and the log showed that the laziness test case was definitely passing. But `:info factorise` showed me that my ghci session was using the `factorise` imported from arithmoi 5.0.0 - I had thought `ghci Primes.hs` would be sufficient to override that, but apparently not. I was finally able to get it to work by giving up on `cabal` and using `stack` instead, with the command line you showed above. (Thanks!)

I've now added a test case for laziness. I stole the same numbers as in the `Integer` case, then tweaked it slightly so that the first prime factor has a nonzero imaginary component.

test-suite/Math/NumberTheory/GaussianIntegersTests.hs
```diff
@@ -14,22 +14,35 @@ module Math.NumberTheory.GaussianIntegersTests
...
+import Control.Monad (zipWithM_)
 import Test.Tasty
 import Test.Tasty.HUnit
 
 import Math.NumberTheory.GaussianIntegers
 import Math.NumberTheory.TestUtils
 
+lazyCases :: [(GaussianInteger, [(GaussianInteger, Int)])]
+lazyCases =
+  [ ( 14145130710
```
Bodigrim: It is better to use big prime p = 4k+1, which cannot be factored by trial division, instead of an even number. For instance, 14145130733 will do. It has non-trivial factors 117058±21037*ι.

cfredric: Done.

Bodigrim:
Great. Just curious: are you obliged to transfer all copyrights to Google?

cfredric:
Yes; adding Google Inc. to the copyrights lines is part of the policy for contributing open source patches. See https://opensource.google.com/docs/patching/.

Bodigrim:
Many thanks! Merged.

Issue #60, opened 2017-07-23
Summatory functions


Bodigrim:
Common difficulty of number theory is that arithmetic function misbehave: their plots oscillate up and down in chaotic fashion. For instance, here is a plot of divisor function from [wiki:](https://en.wikipedia.org/wiki/Divisor_function)
![Divisor function](https://upload.wikimedia.org/wikipedia/commons/thumb/2/20/Divisor.svg/500px-Divisor.svg.png)

The pivot idea of analytical number theory is that instead of studying this jumble, we can study the *summatory* function, which is simply a sum (or an integral) of original function: 

```haskell
summatory :: (Int -> Int) -> Int -> Int 
summatory f n = sum $ map f [1..n]
```

Summatory functions behave well and a wide range of fruitful methods can be applied. E. g., [divisor summatory function](https://en.wikipedia.org/wiki/Divisor_summatory_function). 

The interesting thing is that in many cases summatory function can be computed in sublinear time, faster than O(n). For example, let `tau` denote the divisor function. Then `summatory tau n` is equal to 

```haskell
2 * sum (map (\m -> x `quot` m) [1..sqrtx]) - sqrtx ^ 2
```

where `sqrtx = integerSquareRoot x`. This computation can be completed in O(x^{1/2}) time. There is a systematic way to derive such formulas, which I have explored in [my old paper with a long title](https://arxiv.org/abs/1305.1639).

The proposal is to implement some of these algorithms, to provide users with a DSL to describe functions (in terms of Dirichlet convolution?) and to derive effective summatory formulas from DSL automatically.

CarlEdman:
Very interesting.  I'll dig into that paper later.  Until then I can offer one piece of Haskell code related to the sublinear computation of certain summatory functions over the primes.  Its name is `primeLucy` (in honor of the infamous "Lucy Hedgehog" who first posted the algorithm which this code generalises).  

It takes three parameters: A size parameter `n::Int`, a function `f:: Int -> Integer`, and its summatory function over the integers `sf:: Int -> Integer`.  It returns, in time of the order O(n^0.7) if I recall correctly, a summatory function `g :: Int -> Integer` **over the primes only**.  The summatory function is just a table lookup and hence instant, but is restricted to the parameters [1..sqrt n] and [n/sqrt(n), n/(sqrt(n)-1) .. n].

[...]

PR #77, opened 2017-10-15
Bulk evaluation of arithmetic functions

### LINKED ISSUE #36: sieve design and performance experiments
@ExpHP @Bodigrim  lets talk out concrete ideas / types / experiments / data structures for fancy  nice sieving here :)

Bodigrim:
This patch features a fast bulk evaluation of arithmetic functions. It replaces `FactorSieve`, `TotientSieve`, `CarmichaelSieve` with a unified and extendable interface and makes results freely explorable. 

It is a first step to #60, because it introduces an almost linear algorithm to sum up any arithmetic function. It also resolves #36: new unboxed sieves are about 2x faster and can be used for any function, not only totient and Carmichael ones. It allows to get the smallest prime factor of a number, which was requested in #34 and #67.

### DIFF
--- a/Math/NumberTheory/ArithmeticFunctions/SieveBlock/Unboxed.hs
+++ b/Math/NumberTheory/ArithmeticFunctions/SieveBlock/Unboxed.hs
@@ -1,0 +1,92 @@
+-- |
+-- Module:      Math.NumberTheory.ArithmeticFunctions.SieveBlock.Unboxed
+-- Copyright:   (c) 2017 Andrew Lelechenko
+-- Licence:     MIT
+-- Maintainer:  Andrew Lelechenko <andrew.lelechenko@gmail.com>
+-- Stability:   Provisional
+-- Portability: Non-portable (GHC extensions)
+--
+-- Bulk evaluation of arithmetic functions without factorisation
+-- of arguments.
+--
+
+{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE MagicHash           #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE UnboxedTuples       #-}
+
+module Math.NumberTheory.ArithmeticFunctions.SieveBlock.Unboxed
+  ( SieveBlockConfig(..)
+  , sieveBlockUnboxed
+  ) where
+
+import Control.Monad
+import Control.Monad.ST
+import qualified Data.Vector.Unboxed as V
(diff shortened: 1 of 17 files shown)


Bodigrim:
@ExpHP you've been looking into sieve design (#34 and #36) earlier; would you like to review the patch? If you have enough spare time, of course.

Math/NumberTheory/ArithmeticFunctions/SieveBlock/Unboxed.hs
```diff
@@ -0,0 +1,118 @@
...
+-- >   , sbcBlockLowBound        = 1
+-- >   , sbcBlockLength          = 1000
+-- >   }
+data SieveBlockConfig a = SieveBlockConfig
+  { sbcEmpty                :: a
+    -- ^ value of a function on 1
+  , sbcAppend               :: a -> a -> a
+    -- ^ how to combine values of a function on coprime arguments
+  , sbcFunctionOnPrimePower :: Word -> Word -> a
+    -- ^ how to evaluate a function on prime powers
```
ExpHP: When I first read this it took me a bit to figure out what role each of these functions play. I feel like `sbcEmpty`, `sbcFunctionOnPrimePower`, `sbaAppend` is a more natural order since it puts the base cases together (unless the current order is more natural for currying and I just don't realize it yet)

ExpHP: Actually, it feels odd having these functions and the evaluation range be part of the same object, in general

Math/NumberTheory/ArithmeticFunctions/SieveBlock/Unboxed.hs
```diff
@@ -0,0 +1,118 @@
...
+        logToIx = log (fromIntegral highIndex)
+
+    forM_ ps $ \p -> do
+
+      let p# :: Word#
+          !p'@(W# p#) = intToWord p
+
+          fs = V.generate
+            (floor (logToIx / log (fromIntegral p) + 0.001))
+            (\k -> f p' (intToWord k + 1))
```
ExpHP: seeing floating point numbers appear in number theory sends chills down my spine!  Where does the magic 0.001 come from and how do we know this value works for all inputs?

Bodigrim:
Thank you for thoughtful review and valuable suggestions!

PR #82, opened 2017-10-27
Added prototype for a quadratic integers module. 

blperez01:
Currently has basic functionality for quadratic integers. Still missing factorization capabilities for instances where the ring of integers is a Euclidean domain.

### DIFF
--- a/Math/NumberTheory/QuadraticIntegers.hs
+++ b/Math/NumberTheory/QuadraticIntegers.hs
@@ -1,0 +1,191 @@
+{-# LANGUAGE BangPatterns   #-}
+{-# LANGUAGE DataKinds      #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE PolyKinds #-}
+
+module Math.NumberTheory.QuadraticIntegers (
+  QuadraticInteger(..),
+  Sign(..),
+  discriminant,
+  conjugate,
+  norm,
+  trace,
+  quotRemQ,
+  quotQ,
+  remQ,
+  divModQ,
+  divQ,
+  modQ,
+) where
+
+import GHC.TypeNats.Compat
+import Data.Proxy
+
+infix 5 :+
+
+--------------------------------------------------------------
+-- Type level representation of sign
+  
+data Sign = Pos | Neg
+
+class KnownSign (x :: Sign) where
+  signVal :: proxy x -> Sign
+
+instance KnownSign 'Pos where
+  signVal _ = Pos
+
+instance KnownSign 'Neg where
+  signVal _ = Neg
+
+-- | Takes a quadratic integer and returns its sign as an integer
+signToInt :: forall s d . KnownSign s => QuadraticInteger s d -> Integer
+signToInt _ =
+  case signVal (Proxy :: Proxy s) of
+    Pos -> 1
+    Neg -> -1
+  
+---------------------------------------------------------------
+-- Data definitions for quadratic integers
+
+data QuadraticInteger (sign :: Sign) (d :: Nat) =
+  (:+) { rationalComp :: !Integer, irrationalComp :: !Integer } deriving (Eq)
+
+
+{- If the discriminant of a field is equivalent to 1 mod 4
+   then the basis for the ring of integers is strictly
+   larger than a + bx. Need to represent this other basis -}
+
+data BasisType = Integral | Fractional
+
+
+-- | Takes a quadratic int and returns its basis type
+basisType :: forall s d . (KnownNat d, KnownSign s)
+          => QuadraticInteger s d
+          -> BasisType
+basisType z
+  | disc `mod` 4 == 1 = Fractional
+  | otherwise         = Integral
+  where disc = discriminant z
+
+
+instance Show (QuadraticInteger s d) where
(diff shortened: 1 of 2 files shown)


cartazio:
COol, can you add something to the test suite? And maybe to the benchmark
suite ?

Math/NumberTheory/QuadraticIntegers.hs
```diff
@@ -0,0 +1,191 @@
...
+        where
+            b' = if abs b == 1 then "x" else show (abs b) ++ "*x"
+            op = if b > 0 then "+" else "-"
+            s  = if b > 0 then "" else "-"
+
+instance forall s d . (KnownNat d, KnownSign s) => Num (QuadraticInteger s d) where
+  (+) (a :+ b) (c :+ d) = (a+c) :+ (b+d)
+
+  (*) x@(a :+ b) (c :+ d) =
+    case basisType x of
```
Bodigrim: I am a bit worried that dispatching by `basisType` will take place in runtime instead of compile time. We may introduce a type family like one below, but it will probably fail to compile in older GHCs. 

```haskell
type family GetBasisType (sign :: Sign) (d :: Nat) :: BasisType where
  GetBasisType 'Pos 0 = TypeError ('Text "Discriminant should not be divisible by 4")
  GetBasisType 'Pos 1 = 'Fractional
  GetBasisType 'Pos 2 = 'Integral
  GetBasisType 'Pos 3 = 'Integral
  GetBasisType 'Pos n = GetBasisType 'Pos (n - 4)
  GetBasisType 'Neg 0 = TypeError ('Text "Discriminant should not be divisible by 4")
  GetBasisType 'Neg 1 = 'Integral
  GetBasisType 'Neg 2 = 'Integral
  GetBasisType 'Neg 3 = 'Fractional
  GetBasisType 'Neg n = GetBasisType 'Neg (n - 4)
```

Math/NumberTheory/QuadraticIntegers.hs
```diff
@@ -0,0 +1,191 @@
...
+        | otherwise  = show a ++ op ++ b'
+        where
+            b' = if abs b == 1 then "x" else show (abs b) ++ "*x"
+            op = if b > 0 then "+" else "-"
+            s  = if b > 0 then "" else "-"
+
+instance forall s d . (KnownNat d, KnownSign s) => Num (QuadraticInteger s d) where
+  (+) (a :+ b) (c :+ d) = (a+c) :+ (b+d)
+
+  (*) x@(a :+ b) (c :+ d) =
```
Bodigrim: Please write a test for associativity of multiplication.

Math/NumberTheory/QuadraticIntegers.hs
```diff
@@ -0,0 +1,191 @@
...
+    case basisType x of
+      Integral -> (a*c) + (b*d)*(discriminant x) :+ (a*d + c*b)
+      Fractional -> (a*c + b*d*((discriminant x)-1) `div` 4) :+ (a*d + b*c + b*d)
+
+  abs = undefined
```
Bodigrim: It is better to give a trivial definition than leave `abs` and `signum` undefined. Type class `Num` requires that `x == abs x * signum x`, so one can put `abs x = x` and `signum x = 1 :+ 0`, for instance. 

Another (and better) option is to define `abs` as a map to factor ring by ring units (6 units for Eisenstein numbers, 4 units for Gaussian ones and 2 in other cases). Cf. https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/GaussianIntegers.hs#L68

Math/NumberTheory/QuadraticIntegers.hs
```diff
@@ -0,0 +1,191 @@
...
+
+--------------------------------------------------------------
+-- Division functionality for quadratic integers
+
+-- |Simultaneous 'quot' and 'rem'.
+quotRemQ :: forall s d . (KnownSign s, KnownNat d)
+         => QuadraticInteger s d
+         -> QuadraticInteger s d
+         -> (QuadraticInteger s d, QuadraticInteger s d)
+quotRemQ = divHelper quot
```
Bodigrim: Please add tests for these functions for `QuadraticInteger Neg 4` against their Gaussian counterparts `quotremG` etc.

Math/NumberTheory/QuadraticIntegers.hs
```diff
@@ -0,0 +1,191 @@
...
+-- | Computes the norm of a quadratic integer
+norm :: forall s d . (KnownSign s, KnownNat d)
+     => QuadraticInteger s d
+     -> Integer
+norm x = let (a :+ _) = x * (conjugate x) in a
+
+trace :: forall s d . (KnownSign s, KnownNat d)
+      => QuadraticInteger s d
+      -> Integer
+trace x = let (a :+ _) = x + (conjugate x) in a
```
Bodigrim: Not sure about naming: there is `Debug.Trace.trace` in `base`. Maybe we can use `spur` as old school Germans did? I remember it from my (Russian) books, but I do not know whether it is recognisable by American readers.

Bodigrim:
Great work, thanks for contributing!

Bodigrim:
@blperez01 any chance to complete factorisation capabilities? Looks like the most challenging part of the problem.

Bodigrim:
Closing due to the lack of activity. Anyway thanks for you efforts!

PR #91, opened 2018-01-24
Smooth Numbers module

grandpascorpion:
Functions for generating smooth numbers in various ways.

### DIFF
--- a/Math/NumberTheory/SmoothNumbers.hs
+++ b/Math/NumberTheory/SmoothNumbers.hs
@@ -1,0 +1,114 @@
+-- |
+-- Module:      Math.NumberTheory.SmoothNumbers
+-- Copyright:   (c) 2018 Frederick Schneider
+-- Licence:     MIT
+-- Maintainer:  Frederick Schneider <frederick.schneider2011@gmail.com>
+-- Stability:   Provisional
+-- Portability: Non-portable (GHC extensions)
+--
+-- Functions for generating smooth numbers
+--
+
+module Math.NumberTheory.SmoothNumbers
+  ( smooth
+  , smoothOverSet
+  , smoothOverList
+  , smoothInRange
+  , smoothOverSetInRange
+  , smoothOverListInRange 
+  , smoothInRangeBF
+  , smoothOverSetInRangeBF
+  , smoothOverListInRangeBF
+  ) where
+
+import Data.List (sort, nub)
+import qualified Data.Set as S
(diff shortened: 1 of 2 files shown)


Bodigrim:
Thanks for contributing! Your efforts are much appreciated.

There is a common pattern, which seems applicable here. Namely, to separate validation from actual computations, using a newtype with unexported constructor. 

I suggest to introduce a newtype and following validating routines: 

```haskell
newtype SmoothBasis a = SmoothBasis [a]

fromSet :: S.Set a -> Maybe (SmoothBasis a)
fromSet s = if validU l then Just (SmoothBasis l) else Nothing where l = S.elems s

fromList :: [a] -> Maybe (SmoothBasis a)
fromList l' = if validU l then Just (SmoothBasis l) else Nothing where l = nub $ sort l'
```

Further, we define **total** functions:

```haskell
smoothOver          :: Integral a => SmoothBasis a -> [a]
smoothOverInRange   :: Integral a => SmoothBasis a -> a -> a -> [a]
smoothOverInRangeBF :: Integral a => SmoothBasis a -> a -> a -> [a]
```

There are several advantages:

1. User can easily catch `Nothing`, if his input was invalid. 
   Throwing `error` from internals of the library is much more cumbersome 
   to handle: basically, user must do computations in `IO` monad.

2. User can validate a smooth basis once and use it multiple times. 
   Validation may be quite expensive here.

3. There is no need to have `Set`- and `List`-flavours of each function.

grandpascorpion:
Thanks for the insight, Andrew.  Will try to turn this around shortly.

Best, Fred

Bodigrim:
Awesome! One last suggestion: what UB in `fromSmoothUB` stands for? Could it possibly be renamed to `fromPrimesUpTo` or something similar?

grandpascorpion:
UB means Upper Bound

grandpascorpion:
Cool, thanks for your patience.

By the way, I have released my massive number library:
https://hackage.haskell.org/package/canon

Still early, so there may be some bugs.  The next release will have robust
support of hyperoperations (https://en.wikipedia.org/wiki/Hyperoperation)
so I can finally define Graham's number without blowing up the universe :)

Best, Fred

Bodigrim:
Cool! It would be nice to have some examples of usage directly in haddock documentation.

grandpascorpion:
Great point.  I will try to turn that around today.

grandpascorpion:
Done

PR #101, opened 2018-03-18
Don't report -1 as a factor

### LINKED ISSUE #95: Do not report -1 as a factor
Current `factorise` routines (both `Math.NumberTheory.Primes.Factorisation.Montgomery` and `Math.NumberTheory.GaussianIntegers`) list `-1` as a member of canonical factorisation of negative numbers. This is weird from mathematical perspective (canonical factorisation should consist of primes only) and will eventually be removed, when all primes throughout the library will be represented by newtypes from `Math.NumberTheory.Primes.Types`. 

As a small step towards the goal, one can at least stop reporting `-1` and Gaussian units in factorisations. E. g., `factorise (-1)` should be `[]`. This is a small change, but take care to fix the test suite accordingly.

jhenahan:
Fixes #95 

Not sure if `abs g == abs g'` is the best way to go about this in the property. Feels a little too forgetful, but I'm not sure how else to reformulate the test to get it to pass.

### DIFF
--- a/Math/NumberTheory/GaussianIntegers.hs
+++ b/Math/NumberTheory/GaussianIntegers.hs
@@ -189,10 +189,11 @@ a .^ e
     s = a .^ div e 2
 
 -- |Compute the prime factorisation of a Gaussian integer. This is unique up to units (+/- 1, +/- i).
+-- Unit factors are not included in the result.
 factorise :: GaussianInteger -> [(GaussianInteger, Int)]
 factorise g = helper (Factorisation.factorise $ norm g) g
     where
-    helper [] g' = if g' == 1 then [] else [(g', 1)] -- include the unit, if it isn't 1
+    helper [] _ = []
     helper ((!p, !e) : pt) g' =
         -- For a given prime factor p of the magnitude squared...
         let (!g'', !facs) = if p `mod` 4 == 3
--- a/Math/NumberTheory/Primes/Factorisation/Montgomery.hs
+++ b/Math/NumberTheory/Primes/Factorisation/Montgomery.hs
@@ -78,16 +78,15 @@ import Math.NumberTheory.Primes.Testing.Probabilistic
 import Math.NumberTheory.Unsafe
 import Math.NumberTheory.Utils
 
--- | @'factorise' n@ produces the prime factorisation of @n@, including
---   a factor of @(-1)@ if @n < 0@. @'factorise' 0@ is an error and the
---   factorisation of @1@ is empty. Uses a 'StdGen' produced in an arbitrary
---   manner from the bit-pattern of @n@.
+-- | @'factorise' n@ produces the prime factorisation of @n@. @'factorise' 0@ is
+--   an error and the factorisation of @1@ is empty. Uses a 'StdGen' produced in
+--   an arbitrary manner from the bit-pattern of @n@.
 factorise :: Integer -> [(Integer,Int)]
 factorise n
-    | n < 0     = (-1,1):factorise (-n)
-    | n == 0    = error "0 has no prime factorisation"
-    | n == 1    = []
-    | otherwise = factorise' n
+    | abs n == 1 = []
+    | n < 0      = factorise (-n)
+    | n == 0     = error "0 has no prime factorisation"
+    | otherwise  = factorise' n
 
 -- | Like 'factorise', but without input checking, hence @n > 1@ is required.
 factorise' :: Integer -> [(Integer,Int)]
--- a/test-suite/Math/NumberTheory/GaussianIntegersTests.hs
+++ b/test-suite/Math/NumberTheory/GaussianIntegersTests.hs
@@ -34,7 +34,7 @@ lazyCases =
 factoriseProperty1 :: Integer -> Integer -> Bool
 factoriseProperty1 x y
   =  x == 0 && y == 0
-  || g == g'
+  || abs g == abs g'
   where
     g = x :+ y
     factors = factorise g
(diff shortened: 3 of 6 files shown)


Math/NumberTheory/GaussianIntegers.hs
```diff
@@ -192,7 +192,7 @@ a .^ e
 factorise :: GaussianInteger -> [(GaussianInteger, Int)]
 factorise g = helper (Factorisation.factorise $ norm g) g
     where
-    helper [] g' = if g' == 1 then [] else [(g', 1)] -- include the unit, if it isn't 1
+    helper [] g' = if abs g' == 1 then [] else [(g', 1)] -- include the unit, if it isn't 1
```
Bodigrim: When the first argument is an empty list, `abs g'` is guaranteed to be `1`, so we can just return `[]`. 
In fact, now `helper` do not need to carry the second argument at all.

Math/NumberTheory/Primes/Factorisation/Montgomery.hs
```diff
@@ -84,9 +84,9 @@ import Math.NumberTheory.Utils
 --   manner from the bit-pattern of @n@.
 factorise :: Integer -> [(Integer,Int)]
 factorise n
-    | n < 0     = (-1,1):factorise (-n)
+    | abs n == 1    = []
```
Bodigrim: Please keep code aligned by `=`.

test-suite/Math/NumberTheory/GaussianIntegersTests.hs
```diff
@@ -35,6 +35,7 @@ factoriseProperty1 :: Integer -> Integer -> Bool
 factoriseProperty1 x y
   =  x == 0 && y == 0
   || g == g'
+  || abs g == abs g'
```
Bodigrim: I guess we can safely remove check `g == g'`, because it is a specific instance of `abs g == abs g'`.

Bodigrim:
Thanks for your contribution! Please find review suggestions above.

Math/NumberTheory/GaussianIntegers.hs
jhenahan: I've reduced the base case to

    helper [] _ = []

but I'm less certain about how to rework the other case to remove the need for `g'`, assuming that's what you mean.

jhenahan:
Updated, except as commented above.

Math/NumberTheory/GaussianIntegers.hs
Bodigrim: My fault, it is non-trivial to remove the need for `g'`. Definitely out-of-scope for this PR.

jhenahan:
Haddock comments updated.

Bodigrim:
Excellent! 

There is one remaining area to improve: `Math.NumberTheory.UniqueFactorisation`. Since negative numbers do not cause `-1` as a factor anymore, instances of `UniqueFactorisation` can be simplified. E. g., 

```diff
 instance UniqueFactorisation Int where
   unPrime   = coerce wordToInt
-  factorise m' = if m <= 1
-                then []
-                else map (coerce integerToWord *** intToWord) . F.factorise' . intToInteger $ m
-                  where
-                    m = abs m'
+  factorise = map (coerce integerToWord *** intToWord) . F.factorise . intToInteger
```

Make sure to rebase your branch on latest `master` commit (317c841), please.

jhenahan:
I'll chase down the test failures later on.

Bodigrim:
Just replace `F.factorise'` by 'F.factorise' to fix most of the failures.

Bodigrim:
I looked into two remaining failures. There are several ways to fix them, but the most fundamental one is to change default implementation of `Math.NumberTheory.UniqueFactorisation.isPrime` to handle a special case `isPrime 0 = Nothing` before resorting to `factorise`. This will require enabling `{-# LANGUAGE DefaultSignatures #-}`.

jhenahan:
@Bodigrim Such a definition also implies a `(Num a, Eq a)` constraint on the UFD class, which gets us into trouble in a few other places. It's sufficient to add the special case for `Int` to get the tests to pass, but I'm not sure if you'd prefer a more thorough reformulation to make the special case unnecessary.

Bodigrim:
> Such a definition also implies a (Num a, Eq a) constraint on the UFD class, which gets us into trouble in a few other places.

That is why I suggested enabling `{-# LANGUAGE DefaultSignatures #-}`. One can define `isPrime` this way:

```haskell
  isPrime   :: a -> Maybe (Prime a)

  default isPrime :: (Eq a, Num a) => a -> Maybe (Prime a)
  isPrime 0 = Nothing
  isPrime n = case factorise n of
    [(p, 1)] -> Just p
    _        -> Nothing
```

Then for types, satisfying `(Eq a, Num a)`, `isPrime` will be derived automatically. 

Another option is proceed further with your approach, removing default implementation of `isPrime` entirely and defining it by hand in each of five instances. Less fancy, more robust, so it is a matter of taste.

jhenahan:
I've gone with the `DefaultSignatures` approach for now, but I'd really like to work out a more robust solution to this issue. It might require building out a more expressive typeclass hierarchy to get up to UFDs, though, so it's not at all a blocker for this merge if things otherwise look good.

Bodigrim:
Yeah, it is quite possible that another hierarchy will work better, especially when we get quadratic numbers (#82) merged. `UniqueFactorisation` is still a liquid metal.

Merged, many thanks for your awesome work.

Issue #89, opened 2018-01-02
Wrap output of splitIntoCoprimes into newtype


Bodigrim:
Currently [splitIntoCoprimes](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/GCD.hs#L267) guarantees certain properties of its output, but still returns a plain `[(a, b)]`. Such return type is easily corruptible by other functions. It would be much better to return something like 

```haskell
newtype Coprimes a b = Coprimes (Map a b)
```

The interface should not expose `Coprimes` constructor outside. Here is a quick draft of API:

```haskell
instance Semigroup (Coprimes a b) where ...

instance Monoid (Coprimes a b) where ...

insert :: a -> b -> Coprimes a b -> Coprimes a b 

toList :: Coprimes a b -> [(a, b)]
```

---

Another vaguely related idea is to keep coprimes, which are known to be prime, separately. This can potentially speed up certain computations, especially with `Prefactored`. Like 

```haskell
data Coprimes a b = Coprimes (Map (Prime a) b) (Map a b)
```

cartazio:
Style wise I like to use record fields to document what different positions
in a tuple mean. (One of my fave party tricks Is showing that you can mix
record and fast syntax.  Though that’s not relevant here ;) )

PR #102, opened 2018-04-01
feature/89/coprimes

tau3:
Hello!

I was passing by this project when i saw #89. Hope my PR can help. I'm rather new to haskell, so if my changes somehow violate common code style or best practices, i'll definitlely fix them 😄

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.hi
 *.o
 stack.yaml
+dist
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 *.o
 stack.yaml
 dist
+.*
+!/.gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,3 @@
 *.o
 stack.yaml
 dist
-.*
-!/.gitignore
--- a/Math/NumberTheory/GCD.hs
+++ b/Math/NumberTheory/GCD.hs
@@ -261,18 +261,18 @@ cw32 :: Word32 -> Word32 -> Bool
 cw32 (W32# x#) (W32# y#) = coprimeWord# x# y#
 
 
-newtype Coprimes a b = Coprimes { getValues :: Map.Map a b } deriving (Eq)
+newtype Coprimes a b = Coprimes { unCoprimes :: Map.Map a b } deriving (Eq)
 
 toList :: Coprimes a b -> [(a, b)]
-toList x = Map.assocs $ getValues x
+toList x = Map.assocs $ unCoprimes x
 
 insert :: (Ord a) => a -> b -> Coprimes a b -> Coprimes a b
 insert a b (Coprimes m) = Coprimes (Map.insert a b m)
 
-instance (Ord a) => Semigroup (Coprimes a b) where
-  (Coprimes l) <> (Coprimes r) = Coprimes (Map.union l r)
+instance (Ord a, Ord b) => Semigroup (Coprimes a b) where
+  (Coprimes l) <> (Coprimes r) = Coprimes (Map.unionWith max l r)
 
-instance (Ord a) => Monoid (Coprimes a b) where
+instance (Ord a, Ord b) => Monoid (Coprimes a b) where
   mempty = Coprimes Map.empty
 
 splitIntoCoprimes :: (Integral a, Eq b, Num b) => [(a, b)] -> Coprimes a b
--- a/Math/NumberTheory/GCD.hs
+++ b/Math/NumberTheory/GCD.hs
@@ -26,7 +26,7 @@
 {-# LANGUAGE LambdaCase   #-}
 {-# LANGUAGE MagicHash    #-}
 
-{-# OPTIONS_GHC -fno-warn-unused-imports #-}
+{-# OPTIONS_GHC -fno-warn-unused-imports -fno-warn-unused-top-binds #-}
 
 module Math.NumberTheory.GCD
     ( binaryGCD
@@ -270,14 +270,16 @@ singleton a b = Coprimes (Map.singleton a b)
 toList :: Coprimes a b -> [(a, b)]
 toList x = Map.assocs $ unCoprimes x
 
-insert :: (Ord a) => a -> b -> Coprimes a b -> Coprimes a b
-insert a b (Coprimes m) = Coprimes (Map.insert a b m)
+insert :: (Integral a, Eq b, Num b) => a -> b -> Coprimes a b -> Coprimes a b
+insert a b cs = splitIntoCoprimes ps
+  where ps' = toList cs
+        ps = (a, b) : ps'
 
-instance (Ord a, Integral a, Ord b, Num b) => Semigroup (Coprimes a b) where
+instance (Integral a, Eq b, Num b) => Semigroup (Coprimes a b) where
   (Coprimes l) <> (Coprimes r) = splitIntoCoprimes allTuples
     where allTuples = (Map.assocs l) ++ (Map.assocs r)
 
-instance (Ord a, Integral a, Ord b, Num b) => Monoid (Coprimes a b) where
+instance (Integral a, Eq b, Num b) => Monoid (Coprimes a b) where
   mempty = Coprimes Map.empty
   mappend = (<>)
 
@@ -294,8 +296,7 @@ instance (Ord a, Integral a, Ord b, Num b) => Monoid (Coprimes a b) where
 -- > > splitIntoCoprimes [(360, 1), (210, 1)]
 -- > Coprimes {unCoprimes = fromList [(2,4),(3,3),(5,2),(7,1)]}
 splitIntoCoprimes :: (Integral a, Eq b, Num b) => [(a, b)] -> Coprimes a b
-splitIntoCoprimes xs = foldl (\acc (a,b) -> insert a b acc) (Coprimes Map.empty) tuples
-  where tuples = splitIntoCoprimes' xs
+splitIntoCoprimes xs = Coprimes (Map.fromList $ splitIntoCoprimes' xs)
 
 splitIntoCoprimes' :: (Integral a, Eq b, Num b) => [(a, b)] -> [(a, b)]
 splitIntoCoprimes' xs = if any ((== 0) . fst) ys then [(0, 1)] else go ys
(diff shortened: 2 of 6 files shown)


Bodigrim:
Great! I am on vacation now, I will take a look next weekend.

Math/NumberTheory/GCD.hs
```diff
@@ -26,6 +26,8 @@
 {-# LANGUAGE LambdaCase   #-}
 {-# LANGUAGE MagicHash    #-}
 
+{-# OPTIONS_GHC -Wno-unused-imports #-}
```
Bodigrim: Should be `{-# OPTIONS_GHC -fno-warn-unused-imports #-}`.

tau3:
@Bodigrim it's fixed.

.gitignore
```diff
@@ -1,3 +1,6 @@
 *.hi
 *.o
 stack.yaml
+dist
+.*
+!/.gitignore
```
Bodigrim: It does not look right to me to ignore all dot-files. For instance, this setup ignores `.travis.yml` as well.

Math/NumberTheory/GCD.hs
```diff
@@ -252,6 +260,25 @@ cw16 (W16# x#) (W16# y#) = coprimeWord# x# y#
 cw32 :: Word32 -> Word32 -> Bool
 cw32 (W32# x#) (W32# y#) = coprimeWord# x# y#
 
+
+newtype Coprimes a b = Coprimes { getValues :: Map.Map a b } deriving (Eq)
```
Bodigrim: I'd rather do not use such generic name as `getValues`. I usually follow `newtype Foo a = Foo { unFoo :: a }` convention, so I suggest to name newtype's unwrapper `unCoprimes`.

Bodigrim:
`Coprimes` newtype and all public functions on it should maintain an invariant that elements of `Coprime` are always coprime; this is basically why we wrap it into a newtype. It means that for every `a, b :: Coprime` their union `a <> b` (defined by `Semigroup` instance) must be also valid `Coprime`. Could you please modify your implementation of `Semigroup` to conform with this requirement?

test-suite/Math/NumberTheory/GCDTests.hs
```diff
@@ -86,6 +86,14 @@ toListReturnsCorrectValues :: Assertion
 toListReturnsCorrectValues =
   assertEqual "should be equal" (toList $ splitIntoCoprimes [(140, 1), (165, 1)]) [(5,2),(28,1),(33,1)]
 
+unionReturnsCorrectValues :: Assertion
+unionReturnsCorrectValues =
+  let a = splitIntoCoprimes [(700, 1), (165, 1)] -- [(5,3),(28,1),(33,1)]
+      b = splitIntoCoprimes [(360, 1), (210, 1)] -- [(2,4),(3,3),(5,2),(7,1)]
+      expected = [(2,4),(3,3),(5,3),(7,1),(28,1),(33,1)]
```
Bodigrim: Why is it expected? These numbers are not coprime with each other. I'd expect obtaining `[(2,6),(3,4),(5,5),(7,2),(11,1)]`.

Math/NumberTheory/GCD.hs
```diff
@@ -252,6 +260,25 @@ cw16 (W16# x#) (W16# y#) = coprimeWord# x# y#
...
+toList x = Map.assocs $ unCoprimes x
+
+insert :: (Ord a) => a -> b -> Coprimes a b -> Coprimes a b
+insert a b (Coprimes m) = Coprimes (Map.insert a b m)
+
+instance (Ord a, Ord b) => Semigroup (Coprimes a b) where
+  (Coprimes l) <> (Coprimes r) = Coprimes (Map.unionWith max l r)
+
+instance (Ord a, Ord b) => Monoid (Coprimes a b) where
+  mempty = Coprimes Map.empty
```
Bodigrim: Would not adding `mappend = (<>)` fix compatibility issues with older GHCs?

tau3:
yep, it wasmymistake :(
latest changes didn't help earlier GHCs though

Math/NumberTheory/GCD.hs
```diff
@@ -260,12 +289,12 @@ cw32 (W32# x#) (W32# y#) = coprimeWord# x# y#
 -- having to merge multiplicities of primes, which occurs more than in one
 -- composite factor.
 --
--- > > splitIntoCoprimes [(140, 1), (165, 1)]
+-- > > splitIntoCoprimes' [(140, 1), (165, 1)]
 -- > [(5,2),(28,1),(33,1)]
--- > > splitIntoCoprimes [(360, 1), (210, 1)]
+-- > > splitIntoCoprimes' [(360, 1), (210, 1)]
 -- > [(2,4),(3,3),(5,2),(7,1)]
```
Bodigrim: Please move this comment to `splitIntoCoprimes` and update it accordingly.

Math/NumberTheory/Prefactored.hs
```diff
@@ -100,15 +100,15 @@ fromValue a = Prefactored a [(a, 1)]
 -- > > fromFactors (splitIntoCoprimes [(140, 2), (165, 3)])
 -- > Prefactored {prefValue = 88045650000, prefFactors = [(5, 5), (28, 2), (33, 3)]}
 fromFactors :: Integral a => [(a, Word)] -> Prefactored a
-fromFactors as = Prefactored (product (map (uncurry (^)) as)) (splitIntoCoprimes as)
+fromFactors as = Prefactored (product (map (uncurry (^)) as)) (toList $ splitIntoCoprimes as)
```
Bodigrim: Actually since `prefFactors` keeps "List of pairwise coprime (but not neccesarily prime) factors, accompanied by their multiplicities", we should change its type from `[(a, Word)]` to `Coprimes a Word`. In such case user can analyse contents of `Prefactored` and be sure that it contains what it declares to contain.

And other way around: let `fromFactors` accept `Coprimes a Word`. User should be responsible to provide it with a correct input.

test-suite/Math/NumberTheory/GCDTests.hs
```diff
@@ -82,6 +82,18 @@ splitIntoCoprimesSpecialCase2 :: Assertion
...
+toListReturnsCorrectValues :: Assertion
+toListReturnsCorrectValues =
+  assertEqual "should be equal" (toList $ splitIntoCoprimes [(140, 1), (165, 1)]) [(5,2),(28,1),(33,1)]
+
+unionReturnsCorrectValues :: Assertion
+unionReturnsCorrectValues =
+  let a = splitIntoCoprimes [(700, 1), (165, 1)] -- [(5,3),(28,1),(33,1)]
+      b = splitIntoCoprimes [(360, 1), (210, 1)] -- [(2,4),(3,3),(5,2),(7,1)]
+      expected = [(2,6),(3,4),(5,5),(7,2),(11,1)]
+      actual = toList (a <> b)
```
Bodigrim: Import `Data.Semigroup`: older GHCs does not export `(<>)` from `Prelude`.

tau3:
done.

tau3:
Thank you a lot! 😃 

I risked and attended bonus level
Here is my implementation of `insert` and other fixes. If this implementation doesn't fit, i can revert to straightforward one, and performance improvement could be implemented in a separate PR)

Math/NumberTheory/GCD.hs
```diff
@@ -26,14 +26,23 @@
 {-# LANGUAGE LambdaCase   #-}
 {-# LANGUAGE MagicHash    #-}
 
+{-# OPTIONS_GHC -fno-warn-unused-imports -fno-warn-unused-top-binds #-}
```
Bodigrim: I'd rather export `insert` outside than use `-fno-warn-unused-top-binds`. The latter does not work for GHC 7.10, causing build failure on Travis.

Math/NumberTheory/GCD.hs
```diff
@@ -252,6 +261,31 @@ cw16 (W16# x#) (W16# y#) = coprimeWord# x# y#
 cw32 :: Word32 -> Word32 -> Bool
 cw32 (W32# x#) (W32# y#) = coprimeWord# x# y#
 
+
+newtype Coprimes a b = Coprimes { unCoprimes :: Map.Map a b } deriving (Eq, Show)
+
+singleton :: Ord a => a -> b -> Coprimes a b
+singleton a b = Coprimes (Map.singleton a b)
```
Bodigrim: Actually, I looked into documentation and it appeared that `Map.singleton` does not require `Ord` constraint. Consequently, we do not need it both here and in two occurrences in `Math.NumberTheory.Prefactored`.

Math/NumberTheory/GCD.hs
```diff
@@ -252,6 +261,31 @@ cw16 (W16# x#) (W16# y#) = coprimeWord# x# y#
...
+
+toList :: Coprimes a b -> [(a, b)]
+toList x = Map.assocs $ unCoprimes x
+
+insert :: (Integral a, Bits a, Eq b, Num b) => a -> b -> Coprimes a b -> Coprimes a b
+insert a b cs@(Coprimes m) = if isCoprimeBase
+  then Coprimes (Map.fromList ps)
```
Bodigrim: Would not `Map.insert a b m` be enough? `Map.fromList` is quite expensive.

Math/NumberTheory/GCD.hs
```diff
@@ -252,6 +261,31 @@ cw16 (W16# x#) (W16# y#) = coprimeWord# x# y#
...
+
+newtype Coprimes a b = Coprimes { unCoprimes :: Map.Map a b } deriving (Eq, Show)
+
+singleton :: Ord a => a -> b -> Coprimes a b
+singleton a b = Coprimes (Map.singleton a b)
+
+toList :: Coprimes a b -> [(a, b)]
+toList x = Map.assocs $ unCoprimes x
+
+insert :: (Integral a, Bits a, Eq b, Num b) => a -> b -> Coprimes a b -> Coprimes a b
```
Bodigrim: 1. Why do we need `Bits a` constraint here?
2. Please cover implementation of `insert`with some basic tests

tau3: > Why do we need Bits a constraint here?

`coprime :: (Integral a, Bits a) => a -> a -> Bool` is used in `insert`

tau3:
tests are ready.

Bodigrim:
Merged! Thanks for your contribution and your patience :)

tau3:
woohoo \o/

Issue #87, opened 2018-01-01
Redesign API for modular square roots 


Bodigrim:
Currently `Math.NumberTheory.Moduli.Sqrt` is a fractal of bad design. There are eight functions exposed and **zero** of them are actually suitable to find a modular square root x^2 ≡ n (mod m). User is obliged to factorise a modulo himself and satisfy a list of funny preconditions, stated in comments, but not enforced by types. And if some of preconditions are not satisfied, "the computation may not terminate". 

The proposal is to redesign API from the scratch. There should be a simple function `sqrtMod :: KnownNat m => Mod m -> Maybe (Mod m)`, which actually computes modular square root. Sure, since such function involves factorisation of the modulo at each call, more specific routines should be exposed as well. But:

* Most of preconditions must be encoded in types. If a function expects a prime argument, it must be `Prime Integer` instead of `Integer`.
* Other preconditions (e. g., primes of special form) must be checked in all publicly exposed functions.
* All computations must terminate, even if the input does not match expectations.

Issue #100, opened 2018-03-17
Support Natural on the same level with Int, Word and Integer


Bodigrim:
There are three "first-class" integer (implementing `Integral` instances) types in `arithmoi`: `Int`, `Word` and `Integer`. It means that polymorphic functions `Integral a => ...` has been written, keeping in mind only these three possibilities for `a`. Usually such functions get `{-# SPECIALISE #-}` or/and `{-# RULES #-}` to improve performance. Here is an example from `Math.NumberTheory.Powers.Squares`:

```haskell
integerSquareRoot :: Integral a => a -> a
integerSquareRoot n
  | n < 0       = error "integerSquareRoot: negative argument"
  | otherwise   = integerSquareRoot' n
{-# SPECIALISE integerSquareRoot :: Int -> Int,
                                    Word -> Word,
                                    Integer -> Integer
  #-}

integerSquareRoot' :: Integral a => a -> a
integerSquareRoot' = isqrtA
{-# RULES
"integerSquareRoot'/Int"     integerSquareRoot' = isqrtInt'
"integerSquareRoot'/Word"    integerSquareRoot' = isqrtWord
"integerSquareRoot'/Integer" integerSquareRoot' = isqrtInteger
  #-}
```

Further, such function are tested using `Math.NumberTheory.TestUtils.testIntegralProperty` to run each test thrice with different types.

Nowadays another integer type gains more traction: `Natural`, included in `base` since GHC 8.0. It represents non-negative integers of arbitrary size; `Natural` corresponds to `Integer` as `Word` corresponds to `Int`. We should support it through `arithmoi` on the same level with `Int`, `Word` and `Integer`.

There are three steps:

1. Check that polymorphic functions `Integral a => ...` are actually compatible with `Natural` arguments. This may be not the case, because (in contrast to `Word`) an attempt to create a negative `Natural` throws `Underflow :: ArithException`. 

E. g., `f x = (x - 1) + 2` is a perfectly valid function, when `x` is `Int` or `Integer` and even has an expected semantics for `Word`, but raises an exception on `x = 0 :: Natural`.

One can change `Math.NumberTheory.TestUtils` in the following way and adapt `testIntegralProperty`, `testSameIntegralProperty` and `testIntegral2Property` correspondingly. This should cover majority of cases, and other ones should be dealt manually. 

```diff
 type TestableIntegral wrapper =
-  ( Matrix '[Arbitrary, Show, Serial IO] wrapper '[Int, Word, Integer]
[...]

PR #108, opened 2018-06-08
Add more usable interface to Sqrt module

viviag:
Covers #87 (Not closes since there is still strange internal design)
Possible workaround here:
- [x] Use `PrimeNat`.
- [x] Check if residue is quadratic also outside and reflect it to types - ideally make type like `data Residue a = Quadratic a | NonQuadratic a`. It looks like reflecting terminology to type system is a good approach.

### DIFF
--- a/Math/NumberTheory/GaussianIntegers.hs
+++ b/Math/NumberTheory/GaussianIntegers.hs
@@ -36,7 +36,7 @@ module Math.NumberTheory.GaussianIntegers (
 ) where
 
 import qualified Math.NumberTheory.Moduli as Moduli
-import Math.NumberTheory.Moduli.Sqrt (FieldCharacterictic(..), toFieldCharacteristic)
+import Math.NumberTheory.Moduli.Sqrt (FieldCharacteristic(..), unPrime, toFieldCharacteristic)
 import qualified Math.NumberTheory.Powers as Powers
 import qualified Math.NumberTheory.Primes.Factorisation as Factorisation
 import qualified Math.NumberTheory.Primes.Sieve as Sieve
@@ -139,7 +139,7 @@ primes = [ g
                 else
                     if p == 2
                     then [1 :+ 1]
-                    else let x :+ y = findPrime' (FieldCharacterictic p 1)
+                    else let x :+ y = findPrime' (fromMaybeError "Impossible error" $ toFieldCharacteristic p) -- safe actually
                          in [x :+ y, y :+ x]
          ]
 
@@ -162,9 +162,10 @@ findPrime p
     | otherwise = error "p must be congruent to 3 (mod 4)"
 
 -- |Find a Gaussian integer whose norm is the given prime number.
-findPrime' :: FieldCharacterictic -> GaussianInteger
-findPrime' (FieldCharacterictic p 1) =
-    let (Just c) = Moduli.sqrtModMaybe (-1) (FieldCharacterictic p 1)
+findPrime' :: FieldCharacteristic -> GaussianInteger
+findPrime' (FieldCharacteristic prime 1) =
+    let (Just c) = Moduli.sqrtModMaybe (-1) (FieldCharacteristic prime 1)
+        p = unPrime prime
         k  = Powers.integerSquareRoot p
         bs = [1 .. k]
         asbs = map (\b' -> ((b' * c) `mod` p, b')) bs
--- a/Math/NumberTheory/GaussianIntegers.hs
+++ b/Math/NumberTheory/GaussianIntegers.hs
@@ -30,7 +30,6 @@ module Math.NumberTheory.GaussianIntegers (
     primes,
     gcdG,
     gcdG',
-    findPrime,
     findPrime',
     factorise,
 ) where
@@ -155,13 +154,6 @@ gcdG' g h
     | otherwise = gcdG' h (abs (g `modG` h))
 
 -- |Find a Gaussian integer whose norm is the given prime number.
--- Checks the precondition that p is prime and that p `mod` 4 /= 3.
-findPrime :: Integer -> GaussianInteger
-findPrime p
-    | p == 2 || (p `mod` 4 == 1) = findPrime' (fromMaybeError "p must be prime" $ toFieldCharacteristic p)
-    | otherwise = error "p must be congruent to 3 (mod 4)"
-
--- |Find a Gaussian integer whose norm is the given prime number.
 findPrime' :: FieldCharacteristic -> GaussianInteger
 findPrime' (FieldCharacteristic prime 1) =
     let (Just c) = Moduli.sqrtModMaybe (-1) (FieldCharacteristic prime 1)
@@ -171,6 +163,7 @@ findPrime' (FieldCharacteristic prime 1) =
         asbs = map (\b' -> ((b' * c) `mod` p, b')) bs
         (a, b) = head [ (a', b') | (a', b') <- asbs, a' <= k]
     in a :+ b
+findPrime' (FieldCharacteristic _prime _pow) = error "Not a prime number as argument to findPrime'"
 
 -- |Raise a Gaussian integer to a given power.
 (.^) :: (Integral a) => GaussianInteger -> a -> GaussianInteger
--- a/Math/NumberTheory/GaussianIntegers.hs
+++ b/Math/NumberTheory/GaussianIntegers.hs
@@ -164,7 +164,7 @@ findPrime p
 -- check the precondition.
 findPrime' :: Integer -> GaussianInteger
 findPrime' p =
-    let (Just c) = Moduli.sqrtModP (-1) p
+    let (Just c) = Moduli.sqrtModMaybe (-1) p
         k  = Powers.integerSquareRoot p
         bs = [1 .. k]
         asbs = map (\b' -> ((b' * c) `mod` p, b')) bs
--- a/Math/NumberTheory/GaussianIntegers.hs
+++ b/Math/NumberTheory/GaussianIntegers.hs
@@ -35,11 +35,13 @@ module Math.NumberTheory.GaussianIntegers (
 ) where
 
 import qualified Math.NumberTheory.Moduli as Moduli
-import Math.NumberTheory.Moduli.Sqrt (FieldCharacteristic(..), unPrime, toFieldCharacteristic)
+import Math.NumberTheory.Moduli.Sqrt (FieldCharacteristic(..))
 import qualified Math.NumberTheory.Powers as Powers
+import Math.NumberTheory.Primes.Types (PrimeNat(..))
 import qualified Math.NumberTheory.Primes.Factorisation as Factorisation
 import qualified Math.NumberTheory.Primes.Sieve as Sieve
 import qualified Math.NumberTheory.Primes.Testing as Testing
+import Math.NumberTheory.Utils.FromIntegral (integerToNatural)
 
 infix 6 :+
 infixr 8 .^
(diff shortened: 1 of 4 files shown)


Math/NumberTheory/Moduli/Sqrt.hs
```diff
@@ -13,30 +13,78 @@
...
 
+-- | Extended flag structure to check if square root is computed in finite field.
+data FieldCharacterictic = Invalid
+                         | Prime
+                         | PrimePower Integer Int
+  deriving Show
+
+-- | Check if value is power of some prime.
+checkPrimePower :: Integer -> Maybe (Integer, Int)
+checkPrimePower = headMay . factorise
```
Bodigrim: It should be faster to use `Math.NumberTheory.Powers.General.highestPower` here.

Math/NumberTheory/Moduli/Sqrt.hs
```diff
@@ -13,30 +13,78 @@
...
+  | isPrime p = Prime
+  | otherwise = case checkPrimePower p of
+      Nothing -> Invalid
+      Just (prime,pow) -> PrimePower prime pow
+
+-- * Interface functions with preconditions checking.
+
+sqrtModSharp :: Integer -> Integer -> Integer
+sqrtModSharp n p
+  | p <= 0 = error "Cannot extract sqaure root by negative modulo"
```
Bodigrim: `sqaure` - typo. 
`negative` - should be `non-positive`, because zero is not permitted also.

Math/NumberTheory/Moduli/Sqrt.hs
```diff
@@ -13,30 +13,78 @@
...
+checkPreconditions :: Integer -> FieldCharacterictic
+checkPreconditions p
+  | isPrime p = Prime
+  | otherwise = case checkPrimePower p of
+      Nothing -> Invalid
+      Just (prime,pow) -> PrimePower prime pow
+
+-- * Interface functions with preconditions checking.
+
+sqrtModSharp :: Integer -> Integer -> Integer
```
Bodigrim: What does `Sharp` mean? Like "strict" or "exact"?

Math/NumberTheory/Moduli/Sqrt.hs
```diff
@@ -13,30 +13,78 @@
...
+  | p <= 0 = error "Cannot extract sqaure root by negative modulo"
+  -- Euler's criterion.
+  | isPrime p = if n^((p-1) `div` p) `mod` p == 1
+      then sqrtModP' n p
+      else error "Quadratic nonresidue as argument - in this case use sqrtModList."
+  | otherwise = error "Given characteristic is invalid for the field - cannot compute square roots."
+
+sqrtModList :: Integer -> Integer -> [Integer]
+sqrtModList n p
+  | p <= 0 = error "Cannot extract sqaure root by negative modulo"
```
Bodigrim: Same as above.

Math/NumberTheory/Moduli/Sqrt.hs
```diff
@@ -13,30 +13,78 @@
...
+  | otherwise = case checkPreconditions p of
+      PrimePower prime power -> sqrtModPPList n (prime,power)
+      Prime -> sqrtModPList n p
+      Invalid -> error "Given characteristic is invalid for the field - cannot compute square roots."
+
+sqrtModMaybe :: Integer -> Integer -> Maybe Integer
+sqrtModMaybe n p
+  | p <= 0 = error "Cannot extract sqaure root by negative modulo"
```
Bodigrim: Same as above.

Math/NumberTheory/Moduli/Sqrt.hs
```diff
@@ -30,13 +27,63 @@ import GHC.Integer.GMP.Internals
...
+import Math.NumberTheory.Powers.General (highestPower)
 import Math.NumberTheory.Primes.Sieve (sieveFrom)
+import Math.NumberTheory.Primes.Testing.Probabilistic (isPrime)
 import Math.NumberTheory.Utils (shiftToOddCount, splitOff)
 
+-- | Extended flag structure to check if square root is computed in finite field.
+data FieldCharacterictic = Invalid
+                         | Prime
+                         | PrimePower Integer Int
+  deriving Show
```
Bodigrim: This is a nice approach, but I'd rather separate validation of modulo from actual sqrt computations for the sake of performance. It is quite typical in applications to compute a multitude of square roots modulo same huge prime, and it is better to avoid rerunning `isPrime` on each call. 

One can introduce something like `data FieldCharacteristic = FieldCharacteristic (Prime Integer) Word`, keeping prime number and its power (possibly 1) and change `sqrt*` functions to accept `FieldCharacteristic` directly. 
E. g., `sqrtModList :: Integer -> FieldCharacteristic -> [Integer]`, `sqrtModMaybe :: Integer -> FieldCharacteristic -> Maybe Integer` and so on. 

What do you think about it?

viviag: I agree.

So i see it as
```
data FieldCharacteristic = FieldCharacteristic (Prime Integer) Word

toFieldCharacteristic :: Integer -> FieldCharacteristic
``` in `Math.NumberTheory.Moduli.Class`
And corresponding body of builder, and signatures of `sqrt` functions.

File: Math/NumberTheory/Moduli/Sqrt.hs [hunk 7]
```diff
@@ -33,9 +36,7 @@ import Math.NumberTheory.Primes.Testing.Probabilistic (isPrime)
 import Math.NumberTheory.Utils (shiftToOddCount, splitOff)
 
 -- | Extended flag structure to check if square root is computed in finite field.
-data FieldCharacterictic = Invalid
-                         | Prime
-                         | PrimePower Integer Int
+data FieldCharacterictic = FieldCharacterictic Integer Int
```
Bodigrim: Could we have `Prime Integer` instead of `Integer` here? To make `FieldCharacterictic` correct by construction.

viviag: I'll try

Math/NumberTheory/GaussianIntegers.hs
```diff
@@ -157,14 +158,14 @@ gcdG' g h
 -- Checks the precondition that p is prime and that p `mod` 4 /= 3.
 findPrime :: Integer -> GaussianInteger
```
Bodigrim: Actually no one uses `findPrime`. I suppose we can remove it entirely.

Math/NumberTheory/Moduli/Sqrt.hs
```diff
@@ -10,33 +10,83 @@
...
+unPrime :: PrimeNat -> Integer
+unPrime = coerce naturalToInteger
+
+primeF :: Integer -> PrimeNat
+primeF = coerce integerToNatural
+
+-- | Check if value is power of some prime.
+checkPrimePower :: Integer -> Maybe (Prime Integer, Int)
+checkPrimePower p
+  -- FIXME: @'mapFst'@ function presents in @'Data.Monoid.State'@
```
Bodigrim: Is `mapFst` the same to `Control.Arrow.first`?

viviag: Ah. Yes

test-suite/Math/NumberTheory/Moduli/SqrtTests.hs
```diff
@@ -23,68 +23,75 @@ import Test.Tasty.HUnit
...
 unwrapPP :: (PrimeWrapper Integer, Power Int) -> (Integer, Int)
 unwrapPP (p, Power e) = (unwrapP p, e)
 
+-- FIXME: update names. Maybe not - they describe specific behavior inside interfaces.
+
 -- | Check that 'sqrtMod' is defined iff a quadratic residue exists.
 --   Also check that the result is a solution of input modular equation.
 sqrtModPProperty :: AnySign Integer -> PrimeWrapper Integer -> Bool
-sqrtModPProperty (AnySign n) (unwrapP -> p) = case sqrtModP n p of
+sqrtModPProperty (AnySign n) (unwrapP -> p) = case sqrtModMaybe n (fromJustCustom . toFieldCharacteristic $ p) of
```
Bodigrim: We can avoid partial functions here. Change `unwrapP` to return `Prime Integer` and construct `FieldCharacteristic p 1` directly.

viviag: `Math.NumberTheory.Primes.Types` is hidden by some reason. Should I expose it in order to use `Prime` in tests?

Math/NumberTheory/GaussianIntegers.hs
```diff
@@ -138,7 +138,7 @@ primes = [ g
                 else
                     if p == 2
                     then [1 :+ 1]
-                    else let x :+ y = findPrime' p
+                    else let x :+ y = findPrime' (fromMaybeError "Impossible error" $ toFieldCharacteristic p) -- safe actually
```
Bodigrim: In future `Sieve.primes` will return `[Prime Integer]`, but in meantime we can export `Math.NumberTheory.Primes.Types` and use `FieldCharacteristic (PrimeNat p) 1` directly.

Math/NumberTheory/GaussianIntegers.hs
```diff
@@ -207,7 +201,7 @@ factorise g = helper (Factorisation.factorise $ norm g) g
                 -- otherwise: find a Gaussian prime gp for which `norm gp ==
                 -- p`. Then do trial divisions to find out how many times g' is
                 -- divisible by gp or its conjugate.
-                let gp = findPrime' p
+                let gp = findPrime' (fromMaybeError "p must be prime" $ toFieldCharacteristic p)
```
Bodigrim: Same as above.

Math/NumberTheory/Moduli/Sqrt.hs
```diff
@@ -10,33 +10,85 @@
...
+import Math.NumberTheory.Primes.Testing.Probabilistic (isPrime)
 import Math.NumberTheory.Utils (shiftToOddCount, splitOff)
 
+import Math.NumberTheory.Utils.FromIntegral
+
+-- | Extended flag structure to check if square root is computed in finite field.
+data FieldCharacteristic = FieldCharacteristic (Prime Integer) Int
+  deriving Show
+
+-- | Duplicate functions from @'instance UniqueFactorisation Integer'@. They cannot be imported because of cycle imports.
```
Bodigrim: `Math.NumberTheory.Primes.Types` is already in scope, so we can use `PrimeNat` / `unPrimeNat` directly.

viviag: They're not exported anymore since https://github.com/cartazio/arithmoi/pull/108/commits/33f258515cf029d75a994b44480123923f0b8850
So these are just specified wrappers around `fromIntegral` used to coerce `PrimeNat = PrimeNat Natural` with `Integer` in the module. Since we have `sqrt` functions specified for `Integer` it's ok. Relates #100

viviag:
Seems that approach with `toType` functions is applicable for every restriction e.g. existing wrapper `Prime`. See message of https://github.com/cartazio/arithmoi/pull/108/commits/b0d0c1bd94f67321b08d59c233ca3ae9306e1572

viviag:
Рere we expect `p` is prime and it's checked earlier.
We do not push errors here back and can expect nontermination later if `p` is actually not a prime
`let (Just c) = Moduli.sqrtModMaybe (-1) (FieldCharacteristic (PrimeNat . integerToNatural $ p) 1)`

- [ ] So I will make this function partial - it's better than to make partial `sqrtModMaybe`
I'll leave direct calls to constructor in tests.
- [ ] And it would be good idea not to export constructors at all and name fields in new types. Maybe.

Bodigrim:
Merged! Huge thanks for your contribution.

PR #112, opened 2018-06-14
Implementation of Ramanujan tau-function

### LINKED ISSUE #106: Ramanujan tau-function
Ramanujan tau function is probably the largest omission in `Math.NumberTheory.ArithmeticFunctions.Standard`. It is a multiplicative function, important in theory of modular forms. 

A straightforward approach to its computation is to implement [these formulas](http://www.numbertheory.org/php/tau.html). This will be a nice first step.

More efficient algorithm is given by [Charles 2006](https://link.springer.com/content/pdf/10.1007%2Fs11139-006-6509-y.pdf), but this refinement may be done latter, as a separate task.

b-mehta:
Fixes #106, using the first formula given [here](http://www.numbertheory.org/php/tau.html) and the second from [here](https://en.wikipedia.org/wiki/Ramanujan_tau_function#Ramanujan's_conjectures). The latter could be improved to the binomial summation formula instead, which would prevent lots of unnecessary recalculation for primes to a large power, but this requires an implementation of binomial coefficients as well. 
Tested against OEIS and congruences 1-4 and 8,9 listed [here](https://en.wikipedia.org/wiki/Ramanujan_tau_function#Congruences_for_the_tau_function): note congruence 10 is automatically satisfied by definition. 
An alternative method is proposed [in equation 10](http://mathworld.wolfram.com/TauFunction.html) or in [Charles 2006](https://link.springer.com/content/pdf/10.1007%2Fs11139-006-6509-y.pdf).

### DIFF
--- a/Math/NumberTheory/ArithmeticFunctions/Standard.hs
+++ b/Math/NumberTheory/ArithmeticFunctions/Standard.hs
@@ -153,6 +153,8 @@ jordanHelper pa k = (pa - 1) * pa ^ wordToInt (k - 1)
 ramanujan :: (UniqueFactorisation n, Integral n, Integral m) => n -> m
 ramanujan = runFunction ramanujanA
 
+-- | Calculates the <https://en.wikipedia.org/wiki/Ramanujan_tau_function Ramanujan tau function>
+--   of a positive number @n@, using formulas given <http://www.numbertheory.org/php/tau.html here>
 ramanujanA :: forall n m. (UniqueFactorisation n, Integral n, Integral m) => ArithmeticFunction n m
 ramanujanA = multiplicative $ \((unPrime :: Prime n -> n) -> p) -> ramanujanHelper p
 
--- a/Math/NumberTheory/ArithmeticFunctions/Standard.hs
+++ b/Math/NumberTheory/ArithmeticFunctions/Standard.hs
@@ -26,6 +26,7 @@ module Math.NumberTheory.ArithmeticFunctions.Standard
   , sigma, sigmaA
   , totient, totientA
   , jordan, jordanA
+  , ramanujan, ramanujanA
   , moebius, moebiusA, Moebius(..), runMoebius
   , liouville, liouvilleA
     -- * Additive functions
@@ -148,6 +149,17 @@ jordanHelper pa 1 = pa - 1
 jordanHelper pa 2 = (pa - 1) * pa
 jordanHelper pa k = (pa - 1) * pa ^ wordToInt (k - 1)
 {-# INLINE jordanHelper #-}
+
+ramanujan :: (UniqueFactorisation n, Integral n) => n -> n
+ramanujan = runFunction ramanujanA
+
+ramanujanA :: forall n. (UniqueFactorisation n, Integral n) => ArithmeticFunction n n
+ramanujanA = multiplicative $ \((unPrime :: Prime n -> n) -> p) -> ramanujanHelper p
+
+ramanujanHelper :: (UniqueFactorisation n, Integral n) => n -> Word -> n
+ramanujanHelper _ 0 = 1
+ramanujanHelper p 1 = (65 * (p ^ (11 :: Int) + 1) + 691 * (p ^ (5 :: Int) + 1) - 691 * 252 * sum [sigma 5 k * sigma 5 (p-k) | k <- [1..(p-1)]]) `quot` 756
+ramanujanHelper p k = ramanujanHelper p 1 * ramanujanHelper p (k-1) - (p ^ (11 :: Int)) * ramanujanHelper p (k-2)
 
 moebius :: UniqueFactorisation n => n -> Moebius
 moebius = runFunction moebiusA
--- a/Math/NumberTheory/ArithmeticFunctions/Standard.hs
+++ b/Math/NumberTheory/ArithmeticFunctions/Standard.hs
@@ -150,16 +150,16 @@ jordanHelper pa 2 = (pa - 1) * pa
 jordanHelper pa k = (pa - 1) * pa ^ wordToInt (k - 1)
 {-# INLINE jordanHelper #-}
 
-ramanujan :: (UniqueFactorisation n, Integral n) => n -> n
+ramanujan :: (UniqueFactorisation n, Integral n, Integral m) => n -> m
 ramanujan = runFunction ramanujanA
 
-ramanujanA :: forall n. (UniqueFactorisation n, Integral n) => ArithmeticFunction n n
+ramanujanA :: forall n m. (UniqueFactorisation n, Integral n, Integral m) => ArithmeticFunction n m
 ramanujanA = multiplicative $ \((unPrime :: Prime n -> n) -> p) -> ramanujanHelper p
 
-ramanujanHelper :: (UniqueFactorisation n, Integral n) => n -> Word -> n
+ramanujanHelper :: (UniqueFactorisation n, Integral n, Integral m) => n -> Word -> m
 ramanujanHelper _ 0 = 1
-ramanujanHelper p 1 = (65 * (p ^ (11 :: Int) + 1) + 691 * (p ^ (5 :: Int) + 1) - 691 * 252 * sum [sigma 5 k * sigma 5 (p-k) | k <- [1..(p-1)]]) `quot` 756
-ramanujanHelper p k = ramanujanHelper p 1 * ramanujanHelper p (k-1) - (p ^ (11 :: Int)) * ramanujanHelper p (k-2)
+ramanujanHelper p 1 = (fromIntegral (65 * (p ^ (11 :: Int) + 1) + 691 * (p ^ (5 :: Int) + 1)) - fromIntegral (691 * 252 * sum [sigma 5 k * sigma 5 (p-k) | k <- [1..(p-1)]])) `quot` 756
+ramanujanHelper p k = ramanujanHelper p 1 * ramanujanHelper p (k-1) - (fromIntegral $ p ^ (11 :: Int)) * ramanujanHelper p (k-2)
 
 moebius :: UniqueFactorisation n => n -> Moebius
 moebius = runFunction moebiusA
--- a/Math/NumberTheory/ArithmeticFunctions/Standard.hs
+++ b/Math/NumberTheory/ArithmeticFunctions/Standard.hs
@@ -106,22 +106,22 @@ tau = runFunction tauA
 tauA :: Num a => ArithmeticFunction n a
 tauA = multiplicative $ const (fromIntegral . succ)
 
-sigma :: (UniqueFactorisation n, Integral n) => Word -> n -> n
+sigma :: (UniqueFactorisation n, Integral n, Integral m) => Word -> n -> m
 sigma = runFunction . sigmaA
 
 -- | The sum of the @k@-th powers of (positive) divisors of an argument.
 --
 -- > sigmaA = multiplicative (\p k -> sum $ map (p ^) [0..k])
 -- > sigmaA 0 = tauA
-sigmaA :: forall n. (UniqueFactorisation n, Integral n) => Word -> ArithmeticFunction n n
+sigmaA :: forall n m. (UniqueFactorisation n, Integral n, Integral m) => Word -> ArithmeticFunction n m
 sigmaA 0 = tauA
 sigmaA 1 = multiplicative $ \((unPrime :: Prime n -> n) -> p) -> sigmaHelper p
 sigmaA a = multiplicative $ \((unPrime :: Prime n -> n) -> p) -> sigmaHelper (p ^ wordToInt a)
 
-sigmaHelper :: Integral n => n -> Word -> n
-sigmaHelper pa 1 = pa + 1
-sigmaHelper pa 2 = pa * pa + pa + 1
-sigmaHelper pa k = (pa ^ wordToInt (k + 1) - 1) `quot` (pa - 1)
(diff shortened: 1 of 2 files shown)


Bodigrim:
Super cool! Extra kudos for tests. 

> The latter could be improved to the binomial summation formula instead, which would prevent lots of unnecessary recalculation for primes to a large power, but this requires an implementation of binomial coefficients as well.

Binomial coefficients are implement as [Math.NumberTheory.Recurrencies.Bilinear.binomial](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/Recurrencies/Bilinear.hs#L64).

b-mehta:
Thank you! Ah, I didn't look hard enough clearly for binomial coefficients. I'll have a closer look later.

Bodigrim:
Nice. Could you also please add a simple benchmark to [Math.NumberTheory.ArithmeticFunctionsBench](https://github.com/cartazio/arithmoi/blob/master/benchmark/Math/NumberTheory/ArithmeticFunctionsBench.hs)?

b-mehta:
I've added a benchmark, but the function is significantly slower than the others in the same group so I've used a smaller range.

Bodigrim:
Well done! Taking into account the nature of computations, I suggest to add two pointwise benchmarks:
* Ramanujan tau of prime (around 100000?),
* Ramanujan tau of large power of 2 (around 2^1000?).

Math/NumberTheory/ArithmeticFunctions/Standard.hs
```diff
@@ -149,6 +151,23 @@ jordanHelper pa 2 = (pa - 1) * pa
...
+-- | Calculates the <https://en.wikipedia.org/wiki/Ramanujan_tau_function Ramanujan tau function>
+--   of a positive number @n@, using formulas given <http://www.numbertheory.org/php/tau.html here>
+ramanujanA :: forall n m. (UniqueFactorisation n, Integral n, Integral m) => ArithmeticFunction n m
+ramanujanA = multiplicative $ \((unPrime :: Prime n -> n) -> p) -> ramanujanHelper p
+
+ramanujanHelper :: (UniqueFactorisation n, Integral n, Integral m) => n -> Word -> m
+ramanujanHelper _ 0 = 1
+ramanujanHelper 2 1 = -24
+ramanujanHelper p 1 = (65 * sigmaHelper (p ^ (11 :: Int)) 1 + 691 * sigmaHelper (p ^ (5 :: Int)) 1 - 691 * 252 * 2 * sum [sigma 5 k * sigma 5 (p-k) | k <- [1..(p `quot` 2)]]) `quot` 756
+ramanujanHelper p k = sum [(-1)^j * binomial !! (wordToInt $ k-j) !! (wordToInt $ k - 2*j) * pa^j * tp ^ (k - 2*j) | j <- [0..k `quot` 2]]
```
Bodigrim: I wonder whether this can be improved to avoid exponentiation and random access to a list. Haskell is not clever enough to optimise even `(-1)^j`, not saying about succeeding powers of `pa` and decreasing powers of `tp`. And indexing of a list has linear complexity.

One can write something like 
```
minusOnePowers = cycle [1, -1]
paPowers = iterate (* pa) 1 
tpPowers = reverse $ iterate (* tp ^ 2) (if even k then 1 else tp)
binomials = map (\j -> binomial !! (wordToInt $ k-j) !! (wordToInt $ k - 2*j)) [0 .. k `quot` 2]
```
and zip them altogether 
```
zipWith4 (\a b c d -> a * b * c * d) minusOnePowers paPowers tpPowers binomials
```

This avoids exponentiation, but not random access. I do not know whether there is a clever trick to avoid the latter.

b-mehta: Ah, I (mistakenly) hoped exponentiation would be optimised like that. I'll push with new benchmarks. I couldn't see a nicer way of doing the binomial indexing either - other than that if `j` is in a certain range, computing `k-j` choose `j` is faster.

Bodigrim: 1. `binomial !! (k-j) !! (k-2*j) = binomial !! (k-j) !! j`, where for `j <- [0.. k / 2]` the latter expression looks simpler. 

2. We can always resort to [factorials](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/Recurrencies/Linear.hs#L34). The binomial sequence in question equals to `(k-j)! / (k-2j)! / j!`, which can be built up in a following way:

```haskell
zipWith (\a b c -> a `quot` b `quot` c) xs (skipOdds xs) factorial
  where
    xs = reverse $ take k factorial  
```

`skipOdds` can be copied from [Math.NumberTheory.Zeta](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/Zeta.hs#L41). 

I do not know, whether this factorial approach is faster than indexing

test-suite/Math/NumberTheory/ArithmeticFunctionsTests.hs
```diff
@@ -122,6 +122,32 @@ jordan2Oeis = oeisAssertion "A007434" (jordanA 2)
...
+  where m = ramanujan n - (fromIntegral n) * sigma 9 n :: Integer
+
+-- | ramanujan matches baseline from OEIS.
+ramanujanOeis :: Assertion
+ramanujanOeis = oeisAssertion "A000594" ramanujanA
+  [ 1, -24, 252, -1472, 4830, -6048, -16744, 84480, -113643, -115920
+  , 534612, -370944, -577738, 401856, 1217160, 987136, -6905934, 2727432
+  , 10661420, -7109760, -4219488, -12830688, 18643272, 21288960, -25499225
+  , 13865712, -73279080, 24647168
+  ]
```
Bodigrim: Let's use a longer baseline, e. g., [first 100 values](https://www.wolframalpha.com/input/?i=RamanujanTau%5BRange%5B100%5D%5D). Due to specific nature of computations it is also worth to write tests, comparing [tau(2^k)](https://www.wolframalpha.com/input/?i=RamanujanTau%5B2%5ERange%5B20%5D%5D) and [tau(3^k)](https://www.wolframalpha.com/input/?i=RamanujanTau%5B3%5ERange%5B20%5D%5D) with baselines.

b-mehta:
Apologies, I'll improve tests soon - I also suspect the triangular formula [in equation 10](http://mathworld.wolfram.com/TauFunction.html) will be more efficient for computing values of tau for ranges [1..n] which I'm keen to investigate. For specific values it doesn't seem to create an improvement though.

Bodigrim:
Since Ramanujan tau is multiplicative, one can use [Math.NumberTheory.ArithmeticFunctions.SieveBlock](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/ArithmeticFunctions/SieveBlock.hs) to evaluate it over a range `[1..n]`.

b-mehta:
Sure, but I suspect the triangular number formula may be significantly more efficient

Math/NumberTheory/ArithmeticFunctions/Standard.hs
```diff
@@ -105,22 +107,22 @@ tau = runFunction tauA
 tauA :: Num a => ArithmeticFunction n a
 tauA = multiplicative $ const (fromIntegral . succ)
 
-sigma :: (UniqueFactorisation n, Integral n) => Word -> n -> n
+sigma :: (UniqueFactorisation n, Integral n, Integral m) => Word -> n -> m
```
Bodigrim: I'd rather stick to the previous type signature here and below. Truly, any function `f :: Integral n :: n -> n` can be generalised to `f' :: (Integral n, Integral m) :: n -> m`, but usually there are two reasons to avoid it.

1. Too polymorphic types make type inference feeble. E. g., 

    ```haskell
    sigmaOfSigma :: Integer -> Integer 
    sigmaOfSigma = sigma 1 . sigma 1
    ```

    perfectly compiles for `sigma :: Integral n :: Word -> n -> n`, but fails for `sigma :: (Integral n, Integral m) :: Word -> n -> m`, because GHC does not know which type is emitted by inner `sigma` and consumed by outer one: any `Integral` fits the bill.

2. It is unclear when type conversion actually takes place. There are at least three possibilities: `sigma :: (Integral n, Integral m) :: Word -> n -> m` may

    1. cast its argument `n` to type `m` at the very beginning and then proceed as usual: `sigma a = oldSigma a . fromIntegral`.
    
    2. proceed as usual and cast its result `n` to type `m` at the very end: `sigma a = fromIntegral . oldSigma a`.

    3. cast `n` to `m` during intermediate computations. E. g., in your patch conversion takes place inside `sigmaHelper`.

    If `n ~ Integer` and `m ~ Int`, then each of options above may lead to different results due to overflow will happen at different stages. If `sigma` is less polymorphic (`sigma :: Integral n :: Word -> n -> n`), it becomes clear for an end user where a cast takes place, because he is obliged to insert it manually.

Math/NumberTheory/ArithmeticFunctions/Standard.hs
```diff
@@ -149,6 +151,30 @@ jordanHelper pa 2 = (pa - 1) * pa
...
+ramanujanHelper p 1 = (65 * sigmaHelper (p ^ (11 :: Int)) 1 + 691 * sigmaHelper (p ^ (5 :: Int)) 1 - 691 * 252 * 2 * sum [sigma 5 k * sigma 5 (p-k) | k <- [1..(p `quot` 2)]]) `quot` 756
+ramanujanHelper p k = sum $ zipWith3 (\a b c -> a * b * c) paPowers tpPowers binomials
+  where pa = fromIntegral $ p ^ (11 :: Int)
+        tp = ramanujanHelper p 1
+        paPowers = iterate (* (-pa)) 1
+        binFunc j
+          | 3*j > k   = row !! (wordToInt $ k - 2*j)
+          | otherwise = row !! (wordToInt $ j)
+          where row = binomial !! (wordToInt $ k - j)
+        binomials = map binFunc [0 .. k `quot` 2]
```
Bodigrim: I came up with an idea to avoid random access to the list of binomials. On my machine it is 5x faster in `ArithmeticFunctions/ramanujan prime power` benchmark.

```haskell
        binomials = scanl
          (\acc j -> acc * (k' - 2 * j) * (k' - 2 * j - 1) `quot` (k' - j) `quot` (j + 1))
          1
          [0 .. k' `quot` 2 - 1]
        k' = fromIntegral k 
```

Math/NumberTheory/ArithmeticFunctions/Standard.hs
b-mehta: Sticking to the previous type signature for `sigma` is possible, but for `ramanujanTau` it's unclear to me how this would work - it would cause `ramanujanTau (2 :: Natural)` to create arithmetic underflow, since the output should be negative.  In this way, I'd argue the generalisation of `ramanujanTau` is necessary, unless the underflow behaviour is preferred.

Math/NumberTheory/ArithmeticFunctions/Standard.hs
b-mehta: Lovely! Nicely done.

Math/NumberTheory/ArithmeticFunctions/Standard.hs
Bodigrim: IMHO the most reasonable output type of `ramanujan` is simply `Integer`. Its value grows too rapidly to be confined within bounds of `Int`: `ramanujan 2647 > 2 ^ 63`. And we cannot detect overflow by rapid flip of sign, because `ramanujan` changes sign frequently. So we need a signed arbitrary-precision `Integral` - and basically only `Integer` satisfies these conditions. 

The input of `ramanujan` can be any `Integral`, but I'd rather stick to arbitrary-precision ones. The internals of the algorithm are not opaque for an end user and he cannot predict, whether an overflow may (or is likely to) happen, especially when eleventh powers get involved. That said, the input type should be either `Integer` or `Natural`, but we do not have any standard Haskell class to cover them. For the sake of simplicity, it is not unreasonable to define simply `ramanujan :: Integer -> Integer`.

b-mehta: Sounds good, I've amended.

Math/NumberTheory/ArithmeticFunctions/Standard.hs
```diff
@@ -118,9 +119,9 @@ sigmaA 1 = multiplicative $ \((unPrime :: Prime n -> n) -> p) -> sigmaHelper p
 sigmaA a = multiplicative $ \((unPrime :: Prime n -> n) -> p) -> sigmaHelper (p ^ wordToInt a)
 
 sigmaHelper :: Integral n => n -> Word -> n
-sigmaHelper pa 1 = pa + 1
-sigmaHelper pa 2 = pa * pa + pa + 1
-sigmaHelper pa k = (pa ^ wordToInt (k + 1) - 1) `quot` (pa - 1)
+sigmaHelper pa 1 = fromIntegral $ pa + 1
+sigmaHelper pa 2 = fromIntegral $ pa * pa + pa + 1
+sigmaHelper pa k = fromIntegral $ (pa ^ wordToInt (k + 1) - 1) `quot` (pa - 1)
```
Bodigrim: `fromIntegral` should be redundant now.

Bodigrim:
Looks great! I am keen to merge, if you do not mind. Bulk computations over `[1..n]` may be a subject of a separate branch.

b-mehta:
Sure, sounds good to me.

PR #115, opened 2018-07-28
Add a 'partition' recurrence

### LINKED ISSUE #107: Partition function
[The partition function p(n)](https://en.wikipedia.org/wiki/Partition_(number_theory)#Partition_function) represents the number of possible partitions of a natural number n, which is to say the number of distinct ways of representing n as a sum of natural numbers (with order irrelevant). 

It would be nice to implement the partition function as a self-recurrent sequence `partition :: Num a => [a]` , using [the pentagonal number theorem](https://en.wikipedia.org/wiki/Pentagonal_number_theorem#Relation_with_partitions). The right place for it will be a new module `Math.NumberTheory.Recurrencies`.

rockbmb:
Closes #107.


rockbmb:
Benchmarks for `Math.NumberTheory.partition` as of commit `5ff9a29`:

```
benchmarked Recurrencies/Partition function/partition/!!100
time                 237.4 ns   (203.8 ns .. 261.8 ns)
                     0.923 R²   (0.897 R² .. 0.953 R²)
mean                 268.6 ns   (256.3 ns .. 282.2 ns)
std dev              45.85 ns   (39.81 ns .. 54.31 ns)
variance introduced by outliers: 83% (severely inflated)

benchmarked Recurrencies/Partition function/partition/!!1000
time                 3.403 μs   (3.387 μs .. 3.419 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 3.446 μs   (3.437 μs .. 3.457 μs)
std dev              33.26 ns   (26.85 ns .. 44.39 ns)

benchmarked Recurrencies/Partition function/partition/!!10000
time                 44.19 μs   (43.89 μs .. 44.51 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 44.07 μs   (43.98 μs .. 44.17 μs)
std dev              325.4 ns   (271.7 ns .. 417.7 ns)
```

Math/NumberTheory/Recurrencies.hs
```diff
@@ -0,0 +1,95 @@
...
+pent :: Integral a => a -> a
+pent n = div (3 * (n * n) - n) 2
+
+-- | Infinite list of generalized pentagonal numbers.
+-- Example:
+--
+-- >>> take 10 pents
+-- [0, 1, 2, 5, 7, 12 ,15, 22, 26, 35]
+pents :: Integral a => [a]
+pents = map pent pentIndexes
```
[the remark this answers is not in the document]
rockbmb: The first identity holds, but the second one does not seem to:
```> take 10 $ scanl (\acc n -> acc + 3 * n - 1) 0 [1..]
[0,2,7,15,26,40,57,77,100,126]
```

where the first few generalized pentagonal numbers are `[0, 1, 2, 5, 7, 12, 15, 22, 26, 35, 40, 51, 57, 70, 77, 92, 100, 117, 126]`. I would bet only those with positive indexes are being calculated here.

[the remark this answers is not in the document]
rockbmb:
```
> take 10 $ scanl (\acc n -> acc + 3 * n - 1) (-1) [-1, -2..]
[-1,-5,-12,-22,-35,-51,-70,-92,-117,-145]
```

This means that

```
map pents [0..] = interleave (scanl (\acc n -> acc + 3 * n - 1) 0 [1..])
                                             (scanl (\acc n -> acc + 3 * n - 1) (-1) [-1, -2..])
```

where `interleave :: [a] -> [a] -> [a]` weaves two lists together, element by element, starting with the left argument list.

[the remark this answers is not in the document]
Bodigrim: > The first identity holds, but the second one does not seem to

The second identity looks good to me:

```haskell
> take 10 (map pent [0,-1..])
[0,2,7,15,26,40,57,77,100,126]
> take 10 (scanl (\acc n -> acc + 3 * n - 1) 0 [1..])
[0,2,7,15,26,40,57,77,100,126]
```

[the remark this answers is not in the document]
rockbmb: @Bodigrim My apologies, I misread the left hand side of the second identity...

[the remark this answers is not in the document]
rockbmb: @Bodigrim I just noticed, won't a `Enum a` be necessary for `pents` in any case?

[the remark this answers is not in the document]
Bodigrim: You are right about `Enum a`, my fault.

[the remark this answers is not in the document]
rockbmb: No problem, although I made another mistake in thinking it a problem. `(Enum a, Num a, Ord a)` is necessary for `Integral a`, but not sufficient. I distractedly assumed it was, and that the former implied the latter.

In fact, `Mod m` has instances for `(Enum a, Num a, Ord a)`, but none for `Integral a`, which means having those 3 constraints on `partition` is just fine. Will push a fix.

Math/NumberTheory/Recurrencies.hs
```diff
@@ -0,0 +1,82 @@
...
+-- Examples:
+--
+-- >>> pentagonalSigns [1..5]
+-- [1, 2, -3, -4, 5]
+--
+-- >>> pentagonalSigns [1..6]
+-- [1, 2, -3, -4, 5, 6]
+pentagonalSigns :: Num a => [a] -> [a]
+pentagonalSigns = helper True
+  where
```
Bodigrim: Can we express this manual recursion via some list combinators? Looks like `zipWith` and `cycle` may be applicable.

Math/NumberTheory/Recurrencies.hs
```diff
@@ -0,0 +1,82 @@
...
+pents :: (Enum a, Num a) => [a]
+pents = interleave (scanl (\acc n -> acc + 3 * n - 1) 0 [1..])
+                   (scanl (\acc n -> acc + 3 * n - 2) 1 [2..])
+  where
+    interleave :: [a] -> [a] -> [a]
+    interleave (n : ns) (m : ms) = n : m : interleave ns ms
+    interleave _ _ = []
+
+-- | When calculating the @n@-th partition number @p(n)@ using the sum
+-- @p(n) = p(n-1) + p(n-2) - p(n-5) - p(n-7) + p(11) + ...@, the signs of each
```
Bodigrim: Typo, should be `p(n-11)`.

Math/NumberTheory/Recurrencies.hs
```diff
@@ -0,0 +1,82 @@
...
+    helper :: Num a => Bool -> [a] -> [a]
+    helper b (x : y : ns) | b = x : y : helper b' ns
+                          | otherwise = (-x) : (-y) : helper b' ns
+      where b' = not b
+    helper b (x : _) | b = [x]
+                     | otherwise = [-x]
+    helper _ _ = []
+
+-- @partition !! n@ calculates the @n@-th partition number:
+-- @p(n) = p(n-1) + p(n-2) - p(n-5) - p(n-7) + p(11) + ...@, where @p(0) = 1@
```
Bodigrim: Typo, should be `p(n-11)`.

Math/NumberTheory/Recurrencies.hs
```diff
@@ -0,0 +1,82 @@
...
+-- 
+-- * @tail [0, 1, 2, 5, 7, 12 ,15, 22, 26, 35, ..] == [1, 2, 5, 7, 12 ,15, 22, 26, 35, 40, ..]@.
+-- * @takeWhile (\m -> 10 - m >= 0) == [1, 2, 5, 7]@.
+-- * @map (\m -> dict ! fromIntegral (10 - m)) [1, 2, 5, 7] == [dict ! 9, dict ! 8, dict ! 5, dict ! 3] == [30, 22, 7, 3]@
+-- * @pentagonalSigns [30, 22, 7, 3] == [30, 22, 7, 3] == [30, 22, -7, -3]@
+-- * @sum [30, 22, -7, -3] == 42@
+--
+-- Note: @tail@ is applied to @pents@ because otherwise the calculation of
+-- @p(n)@ would involve a duplicated @p(n-1)@ term (see the above example).
+partition :: forall a . (Enum a, Num a, Ord a) => [a]
```
Bodigrim: Please write a test checking that, for example, `partition :: [Mod 10]` coincides with `partition :: [Integer]` modulo 10.

Math/NumberTheory/Recurrencies.hs
rockbmb: Got it.

rockbmb:
@Bodigrim can you try doing `take 10 (Math.NumberTheory.Recurrencies.partition) :: [Math.NumberTheory.Moduli.Mod 10]` in a `cabal new-repl` session?

It calculates the head of the list, and that's as far as it goes. Consumed 12GB of RAM before I killed the session.

test-suite/Math/NumberTheory/Recurrencies/PentagonalTests.hs
```diff
@@ -48,10 +53,30 @@ partitionProperty1 (Positive n) =
...
+partitionProperty2 :: NonNegative Integer -> Nat -> Bool
+partitionProperty2 (NonNegative m) n = driver
+  where
+    m' = fromIntegral m
+    nat = natVal @n
+    p :: KnownNat p => p
+    p = case someNatVal n of
+        SomeNat (_ :: Proxy t) -> t
+    driver :: KnownNat p => Bool
+    driver = take m' (partition :: KnownNat n => [Mod p]) ==
```
rockbmb: @Bodigrim I haven't made this work yet, but the idea is present and this and the next lines. I'm not sure how I'll reconcile the `[SomeMod]` produced by helper with the `[Mod n]` that `partition` return.

Bodigrim: We can either apply `fromInteger :: Integer -> Mod p` to `partition :: [Integer]` and compare with `partition :: [Mod p]`, or on contrary extract integers from `partition :: [Mod p]` by `getVal` and compare with `map (flip mod p) partition :: [Integer]`. 

Or wrap everything up to `SomeMod`, employing its `Eq` instance, applying `SomeMod` constructor to `Mod p` and `flip modulo p` to `Integer`.

Math/NumberTheory/Recurrencies/Pentagonal.hs
```diff
@@ -0,0 +1,74 @@
...
+-- @p(n)@ would involve a duplicated @p(n-1)@ term (see the above example).
+partition :: forall a . (Enum a, Num a, Ord a) => [a]
+partition = 1 : go (M.insert 0 1 mempty) 1
+  where
+    go :: (Enum a, Num a, Ord a) => M.Map a a -> a -> [a]
+    go dict !n =
+        let n' = (sum .
+                 pentagonalSigns .
+                 map (\m -> dict M.! (n - m)) .
+                 takeWhile (\m -> n - m >= 0) .
```
Bodigrim: > @Bodigrim can you try doing take 10 (Math.NumberTheory.Recurrencies.partition) :: [Math.NumberTheory.Moduli.Mod 10] in a cabal new-repl session?
>
> It calculates the head of the list, and that's as far as it goes. Consumed 12GB of RAM before I killed the session.

I can reproduce the issue. AFAIU it happens because the condition `n - m >= 0 :: [Mod m]` is always true in modular arithmetic. Can we actually employ `Map Int a` or even `IntMap a` for caching here?

rockbmb: >Can we actually employ Map Int a or even IntMap a for caching here?
Are you asking if it is feasible, or if it is at all what we want to do?

Bodigrim: My impression is that it is possible and desirable. `maxBound :: Int` is big enough to store indices, so no worries about possible overflow here.

test-suite/Math/NumberTheory/Recurrencies/PentagonalTests.hs
rockbmb: >or on contrary extract integers from partition :: [Mod p] by getVal and compare with map (flip mod p) partition :: [Integer]

I'll need to somehow reflect the `Natural` passed to the test as an argument into a `Nat`. `someNatVal` does `:: Natural -> SomeNat`, but I'm not sure how to do `SomeNat -> Nat`, but at the type level.

Bodigrim: Actually you cannot pass any term of type `Nat` to `partitionProperty2`, because `Nat` is an uninhabited type (but "inhabited" kind). So we should go other way around: pass `Natural` to `partitionProperty` and reflect it on type level as `Nat`, e. g., following pattern from [modulo](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/Moduli/Class.hs#L216).

rockbmb: I did it just before you posted this, now I'm fixing the fact that the test will fail because `smallcheck:                     spec: internal error: Unable to commit 1048576 bytes of memory`. This is related to below's comments about using `Map Int a` in `partition`. I'll commit the corrected test for now.

Math/NumberTheory/Recurrencies/Pentagonal.hs
rockbmb: Even after using `Map Int a`, doing `partition :: [Mod 10]` won't calculate more than 7 terms. I think it's because of the following line, `tail) pents`. By asking for `partition` with type `[Mod 10]`, what happens is that `pents` also has that type, and this spoils the remainder of the calculation. The following `takeWhile` won't stop where it's supposed to, and so on.

I'll try giving `pents` type `[Int]` and see what happens.

rockbmb:
@Bodigrim Travis CI's Haddock builds tell me:
```
50% (  2 /  4) in 'Math.NumberTheory.Recurrencies.Pentagonal'
  Missing documentation for:
    Module header
```

The module header part is the one that has the `Copyright:/Maintainer` fields, right? Is there anything in particular I need to put there?

rockbmb:
Benchmark with `Map`:
```
benchmarked Recurrencies/Pentagonal/Partition function/partition/!!1000
time                 8.449 μs   (7.654 μs .. 9.516 μs)
                     0.919 R²   (0.883 R² .. 0.952 R²)
mean                 8.802 μs   (8.394 μs .. 9.332 μs)
std dev              1.450 μs   (1.187 μs .. 1.838 μs)
variance introduced by outliers: 81% (severely inflated)

benchmarked Recurrencies/Pentagonal/Partition function/partition/!!10000
time                 74.68 μs   (66.34 μs .. 84.05 μs)
                     0.888 R²   (0.822 R² .. 0.943 R²)
mean                 97.04 μs   (91.75 μs .. 106.8 μs)
std dev              21.70 μs   (15.41 μs .. 34.40 μs)
variance introduced by outliers: 89% (severely inflated)

benchmarked Recurrencies/Pentagonal/Partition function/partition/!!100000
time                 3.629 ms   (3.455 ms .. 3.833 ms)
                     0.981 R²   (0.961 R² .. 0.994 R²)
mean                 3.370 ms   (3.294 ms .. 3.461 ms)
std dev              241.9 μs   (184.1 μs .. 371.2 μs)
variance introduced by outliers: 37% (moderately inflated)

Benchmark criterion: FINISH
```

Benchmark with `IntMap`:
```
benchmarked Recurrencies/Pentagonal/Partition function/partition/!!1000
time                 4.951 μs   (4.893 μs .. 5.000 μs)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 4.969 μs   (4.941 μs .. 5.009 μs)
std dev              114.5 ns   (89.21 ns .. 157.1 ns)

benchmarked Recurrencies/Pentagonal/Partition function/partition/!!10000
time                 59.54 μs   (58.64 μs .. 60.56 μs)
                     0.998 R²   (0.996 R² .. 1.000 R²)
mean                 60.45 μs   (60.03 μs .. 61.42 μs)
std dev              1.987 μs   (1.065 μs .. 3.635 μs)
variance introduced by outliers: 16% (moderately inflated)

benchmarked Recurrencies/Pentagonal/Partition function/partition/!!100000
time                 3.019 ms   (2.993 ms .. 3.041 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 3.023 ms   (3.009 ms .. 3.035 ms)
std dev              38.90 μs   (29.73 μs .. 56.17 μs)

Benchmark criterion: FINISH
```

rockbmb:
@Bodigrim  Inside `partition`, do you have any preference for

```
go :: forall a . (Enum a, Num a, Ord a) => IM.IntMap a -> Int -> [a]
    go dict !n =
        let intN = fromEnum n
            n' = (sum .
                  pentagonalSigns .
                  map (\m -> dict IM.! (intN - m)) .
                  takeWhile (\m -> intN >= m) .
                  tail) (pents :: [Int])
            dict' = IM.insert intN n' dict
        in n' : go dict' (n+1)
```

or

```
go :: forall a . (Enum a, Num a, Ord a) => IM.IntMap a -> [Int] -> [a]
    go dict (n : ns) =
        let intN = fromEnum n
            n' = (sum .
                  pentagonalSigns .
                  map (\m -> dict IM.! (intN - m)) .
                  takeWhile (\m -> intN >= m) .
                  tail) (pents :: [Int])
            dict' = IM.insert intN n' dict
        in n' : go dict' ns
    go _ _ = []
```
?

Math/NumberTheory/Recurrencies/Pentagonal.hs
```diff
@@ -0,0 +1,75 @@
...
+-- 
+-- * @tail [0, 1, 2, 5, 7, 12 ,15, 22, 26, 35, ..] == [1, 2, 5, 7, 12 ,15, 22, 26, 35, 40, ..]@.
+-- * @takeWhile (\m -> 10 - m >= 0) == [1, 2, 5, 7]@.
+-- * @map (\m -> dict ! fromIntegral (10 - m)) [1, 2, 5, 7] == [dict ! 9, dict ! 8, dict ! 5, dict ! 3] == [30, 22, 7, 3]@
+-- * @pentagonalSigns [30, 22, 7, 3] == [30, 22, 7, 3] == [30, 22, -7, -3]@
+-- * @sum [30, 22, -7, -3] == 42@
+--
+-- Note: @tail@ is applied to @pents@ because otherwise the calculation of
+-- @p(n)@ would involve a duplicated @p(n-1)@ term (see the above example).
+partition :: forall a . (Enum a, Num a, Ord a) => [a]
```
Bodigrim: `Enum` and `Ord` constraints seem to be redundant.

Math/NumberTheory/Recurrencies/Pentagonal.hs
```diff
@@ -0,0 +1,75 @@
...
+partition = 1 : go (IM.singleton 0 1) [1..]
+  where
+    go :: forall a . (Enum a, Num a, Ord a) => IM.IntMap a -> [Int] -> [a]
```
Bodigrim: `Enum` and `Ord` constraints seem to be redundant.

Math/NumberTheory/Recurrencies/Pentagonal.hs
```diff
@@ -0,0 +1,75 @@
...
+    go dict (n : ns) =
+        let intN = fromEnum n
```
Bodigrim: AFAIU both `n` and `intN` have type `Int`, so this cast is redundant.

test-suite/Math/NumberTheory/Recurrencies/PentagonalTests.hs
```diff
@@ -0,0 +1,80 @@
...
+
+-- | Check that @p(n) = p(n-1) + p(n-2) - p(n-5) - p(n-7) + p(n-11) + ...@,
+-- where @p(x) = 0@ for any negative integer.
+partitionProperty1 :: Positive Int -> Bool
+partitionProperty1 (Positive n) =
+    partition' n == (sum .
+                     pentagonalSigns .
+                     map (\m -> partition' (n - m)) .
+                     takeWhile (\m -> n - m >= 0) .
+                     tail $ pents)
```
Bodigrim: If your intention was to compare `partition` against the reference implementation, it is worth to copy-paste reference definitions of `pents` and `pentagonalSigns` to tests as well. Otherwise, if one break, for instance, `pents`, this test would not be able to catch it.

It is also worth to check that first, like, 20 values of `partition` match the baseline. Brutal, but practical.

Bodigrim:
> @Bodigrim Inside partition, do you have any preference for...

I'd prefer the first snippet.

test-suite/Math/NumberTheory/Recurrencies/PentagonalTests.hs
rockbmb: @Bodigrim it is probably not necessary to copy-paste `pents` as `pentagonalNumbersProperty1` in this module checks that the sequence is correct.

rockbmb:
@Bodigrim The PR should be complete now.

Bodigrim:
Merged. Thanks for your efforts and contribution!

PR #116, opened 2018-07-31
Speed up factorisation of Gaussian primes

### LINKED ISSUE #111: Find Gaussian prime with a given norm in effective way
There are three kinds of Gaussian primes (see [Math.NumberTheory.GaussianIntegers.primes](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/GaussianIntegers.hs#L133)): 
* 1 + i, 
* p + 0i, where p is an integer prime and p = 4k+3,
* a + bi, where a^2+b^2 is an integer prime and p = 4k+1.

In the third case it is non-trivial how to find a and b, satisfying a^2+b2 = p for a given p. Currently [Math.NumberTheory.GaussianIntegers.findPrime'](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/GaussianIntegers.hs#L165) does it via brute force. 

We should rather employ the Hermite-Serret algorithm. Resources:
https://math.stackexchange.com/a/5883
http://www.ams.org/journals/mcom/1972-26-120/S0025-5718-1972-0314745-6/S0025-5718-1972-0314745-6.pdf

Bodigrim:
This PR resolves #111. 
@cfredric would you be interested to review?


Math/NumberTheory/GaussianIntegers.hs
```diff
@@ -155,15 +158,20 @@ gcdG' g h
...
+-- <http://www.ams.org/journals/mcom/1972-26-120/S0025-5718-1972-0314745-6/S0025-5718-1972-0314745-6.pdf Hermite-Serret algorithm>.
 findPrime' :: Integer -> GaussianInteger
-findPrime' p =
-    let (Just c) = Moduli.sqrtModMaybe (-1) (FieldCharacteristic (PrimeNat . integerToNatural $ p) 1)
-        k  = Powers.integerSquareRoot p
-        bs = [1 .. k]
-        asbs = map (\b' -> ((b' * c) `mod` p, b')) bs
-        (a, b) = head [ (a', b') | (a', b') <- asbs, a' <= k]
-    in a :+ b
+findPrime' p = case Moduli.sqrtModMaybe (-1) (FieldCharacteristic (PrimeNat . integerToNatural $ p) 1) of
```
cfredric: Not related to this PR, but it seems odd to have `findPrime'` when there is no `findPrime`. Is it worth renaming to `findPrime`, and keeping `findPrime'` around as a deprecated name?

Math/NumberTheory/GaussianIntegers.hs
```diff
@@ -181,50 +189,61 @@ a .^ e
     | otherwise = a * a .^ (e - 1)
     where
     s = a .^ div e 2
+{-# DEPRECATED (.^) "Use (^) instead." #-}
 
 -- |Compute the prime factorisation of a Gaussian integer. This is unique up to units (+/- 1, +/- i).
 -- Unit factors are not included in the result.
 factorise :: GaussianInteger -> [(GaussianInteger, Int)]
-factorise g = helper (Factorisation.factorise $ norm g) g
+factorise g = concat $ snd $ mapAccumL go g (Factorisation.factorise $ norm g)
```
cfredric: Is there a source for this algorithm (that might be worth citing from the doc comment)?

test-suite/Math/NumberTheory/GaussianIntegersTests.hs
```diff
@@ -15,80 +15,143 @@ module Math.NumberTheory.GaussianIntegersTests
...
+gcdGProperty1 z1 z2
+  = z == 0
+  || z1 `remG` z == 0 && z2 `remG` z == 0 && z == abs z
+  where
+    z = gcdG z1 z2
+
+gcdGProperty2 :: GaussianInteger -> GaussianInteger -> GaussianInteger -> Bool
+gcdGProperty2 z z1 z2
+  = z == 0
+  || gcdG z1' z2' `remG` z == 0
```
cfredric: Is this a bug? How can the GCD be zero? (Or am I reading this wrong?)

rockbmb: I don't think it's asking of the `gcd` is zero, but if the remainder of the division of that `gcd` by `z` is 0.

Bodigrim: It is convenient to set `gcd 0 0` equal to `0`.

Math/NumberTheory/GaussianIntegers.hs
Bodigrim: Good point. Will do.

Math/NumberTheory/GaussianIntegers.hs
Bodigrim: The idea is to avoid repetitive complex divisions with the same divisor, because they unnecessary recalculate its norm. I'll refactor code to make it more palatable.

cfredric: Ah, I see. I think I was thrown off by the extra units that entered into the equations (to make the math work out nicer).

test-suite/Math/NumberTheory/GaussianIntegersTests.hs
cfredric: I see, I was parsing the code incorrectly.

cfredric:
LGTM, thanks!

Bodigrim:
@cfredric Thanks for review!

PR #121, opened 2018-08-09
[#118] Add Eisenstein integers module

### LINKED ISSUE #118: Eisenstein integers
My feeling is that currently #73 is too broad and too vague. We should rather explore more special cases to hone API before taking a stab at the general problem. 

We have already implemented the simplest quadratic ring, which is the ring of [Gaussian integers](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/GaussianIntegers.hs). The next step is to handle the ring of Eisenstein integers. 

The API may follow the outline of [Math.NumberTheory.GaussianIntegers](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/GaussianIntegers.hs). Required mathematical background can be found here:
* https://en.wikipedia.org/wiki/Eisenstein_integer
* https://en.wikipedia.org/wiki/Eisenstein_prime

rockbmb:
This will close #118, but at the moment it's still a work in progress.

EDIT:

- [x] Fix `divModE`, and write more extensive tests for Euclidean division (in the commutative ring of Eisenstein integers).
- [x] Write an `isPrime :: EisensteinInteger -> Bool` function.
- [x] Add a `gcdE :: EisensteinInteger -> EisensteinInteger -> EisensteinInteger ` function.
- [x] Write a `factorize :: EisensteinInteger -> [(EisensteinInteger, Int)]` function.
- [x] Add a way to generate all Eisenstein primes, similar to what is done with `GaussianInteger`s. This has been split from the `isPrime` bulletpoint as it will not be a trivial task.

All of these features will require
- [x] Tests
 and 
- [x] Benchmarks.

 There are other functions of use to be added, but these are the most important.


Math/NumberTheory/EisensteinIntegers.hs
```diff
@@ -0,0 +1,68 @@
...
+
+instance Num EisensteinInteger where
+    (+) (a :+ b) (c :+ d) = (a + c) :+ (b + d)
+    (*) (a :+ b) (c :+ d) = (a * c - b * d) :+ (b * c + a * d - b * d)
+    -- An Eisenstein integer @a :+ b@, with @a, b@ integers, can we written as
+    -- @(2*a - b) / 2 + ((b * sqrt 3) * ι) / 2@, but this number is in the
+    -- same quadrant as @(2*a - b) / 2 + (b * ι) / 2@, and this one in the
+    -- same as @(2*a - b) + b * ι@. Divisions or floating points are not
+    -- necessary.
+    abs z@(x :+ y) = abs' (2*x - y) x
```
rockbmb: @Bodigrim I just realized this is wrong. First of all, what one take the absolute value of an Eisenstein integer, `abs n` to be? Is it an E. integer `n'` with the same magnitude such that `(signum n) * (abs n) == n'`?

With Gaussian integers, there are four units, one for each quadrant of the complex plane, so `signum` always returns one of those.
However, the commutative ring of Eisenstein integers has 6 units, `{1, -1, ω, -ω, ω², -ω²}`, one cannot assign each to a regular Cartesian quadrant. Do we divide the complex plane into 6 parts when dealing with Eisenstein integers?

Bodigrim: Exactly, there are 6 sextants, 60 degrees each. `abs` should put its argument into the first one.

Math/NumberTheory/EisensteinIntegers.hs
```diff
@@ -44,19 +56,16 @@ instance Show EisensteinInteger where
...
     (+) (a :+ b) (c :+ d) = (a + c) :+ (b + d)
     (*) (a :+ b) (c :+ d) = (a * c - b * d) :+ (b * c + a * d - b * d)
-    -- An Eisenstein integer @a :+ b@, with @a, b@ integers, can we written as
-    -- @(2*a - b) / 2 + ((b * sqrt 3) * ι) / 2@, but this number is in the
-    -- same quadrant as @(2*a - b) / 2 + (b * ι) / 2@, and this one in the
-    -- same as @(2*a - b) + b * ι@. Divisions or floating points are not
-    -- necessary.
-    abs z@(x :+ y) = abs' (2*x - y) x
+    abs z@(a :+ b)
+        | a == 0 && b == 0                                  = z            -- origin
```
rockbmb: @Bodigrim There is probably a better way to do this. I did not want to use `phase` here because it brings `Double`'s precision issues as the numbers get larger.

`abs $ 10^n :+ (10^n - 1)` will probably give the wrong result for some large `n`, but for now it works. I'll think of a way to do this check without phases later.

rockbmb: @Bodigrim I derived those bounds as well, but I got wrong results the first time because I was multiplying by the wrong unit (i.e. perfoming an incorrect rotation in the plane), so I decided to make it work first with `Data.Complex.phase` and then think about it some more.

Bodigrim:
> @Bodigrim unrelated to the issue, isn't it about time the package is moved away from GHC==7.10.3?

The latest release still supports GHC 7.8 and the upcoming one will be the first which doesn't. I am not keen to drop support of GHC 7.10 as well that soon. Is there anything particularly annoying in 7.10 from your point of view?

Bodigrim:
IMHO with a new, half-year release cycle of GHC it is preferable to support more than three last versions, as long as it does not cause too much pain. I do not like CPP pragmas and feel excited each time when I can get rid of them, of course :) 

Ubuntu 16.04 LTS was shipped with GHC 7.10 and I have plenty of them. Of course, nowadays we rarely use GHC from system distribution, but I mean, 7.10 was not that long ago.

Alternatively, as a last resort, one can expose new functionality for newer GHCs only: 

```
  if impl(ghc >= 8)
    exposed-modules:
      Math.NumberTheory.BrandNewStuff
```

(I wrestle against temptation to use this trick to implement ideas described in #71.)

rockbmb:
Status update: right now, the division algorithm for `EisensteinInteger`s isn't working (fails a simple test), but the plan for the PR is as follows:

- [ ] Fix `divModE`, and write more extensive tests for Euclidean division (in the commutative ring of Eisenstein integers).

- [ ] Write an `isPrime :: EisensteinInteger -> Bool` function, along with a way to generate all Eisenstein primes, similar to what is done with `GaussianInteger`s.

- [ ] Add a `gcdE :: EisensteinInteger -> EisensteinInteger -> EisensteinInteger ` function.

- [ ] Write a `factorize :: EisensteinInteger -> [(EisensteinInteger, Int)]` function.

All of these features will require tests.

Bodigrim:
Working on #129, I discovered that the general case of quadratic modular equations is harder than I thought. But there is a simple method, when a coefficient at x^2 is 1. Here is how it works for Eisenstein primes. 

---

Each prime of form 3n+1 is actually of form 6k+1.
We have (z+3k)^2 = z^2 + 6kz + 9k^2 = z^2 - z + 9k^2 (mod 6k+1).

Our goal is to solve z^2 - z + 1 = 0 (mod 6k+1). We have:
z^2 - z + 9k^2 = 9k^2 - 1 (mod 6k+1)
(z+3k)^2 = 9k^2-1 (mod 6k+1)
z+3k = sqrtMod(9k^2-1) 
z = sqrtMod(9k^2-1) - 3k

For example, let p = 7, then k = 1. Square root of 9\*1^2-1 modulo 7 is 1. 
And z = 1 - 3*1 = -2 = 5 (mod 7). Truly, norm(5 :+ 1) = 25 - 5 + 1 = 21 = 0 (mod 7).

rockbmb:
@Bodigrim sorry, where did this

> We have `(z+3k)^2 = z^2 + 6kz + 9k^2 = z^2 - z + 9k^2 (mod 6k+1)`.

come from?

Bodigrim:
I update the previous comment. Is it more clear now?

rockbmb:
@Bodigrim Yes, but it was clear from the start. What I mean is, was that fact just stated to be used afterwards?

Math/NumberTheory/EisensteinIntegers.hs
```diff
@@ -127,4 +131,17 @@ conjugate (a :+ b) = (a - b) :+ (-b)
...
+-- | Checks if a given @EisensteinInteger@ is prime. @EisensteinInteger@s
+-- whose norm is a prime congruent to @0@ or @1@ modulo 3 are prime.
+-- See <http://thekeep.eiu.edu/theses/2467 Bandara, Sarada, "An Exposition of the Eisenstein Integers" (2016)>,
+-- page 12.
+isPrime :: EisensteinInteger -> Bool
+isPrime e@(a :+ b)
+    | e == 0                             = False
+    | b == 0 && a `mod` 3 == 2           = Testing.isPrime a
+    | a == 0 && b `mod` 3 == 2           = Testing.isPrime b
+    | nE `mod` 3 == 0 || nE `mod` 3 == 1 = Testing.isPrime nE
```
rockbmb: I don't know how I didn't notice what I wrote here, but this is clearly wrong. `nE 'mod' 3 == 0` will never happen for a prime.

Bodigrim:
> What I mean is, was that fact just stated to be used afterwards?
Yes, just used afterwards.

Math/NumberTheory/EisensteinIntegers.hs
```diff
@@ -0,0 +1,345 @@
...
+        | b >= a && a > 0                                   = (-ω) * z     -- second sextant: π/3 ≤ Arg(η) < 2π/3
+        | b > 0 && 0 >= a                                   = (-1 - ω) * z -- third sextant: 2π/3 ≤ Arg(η) < π
+        | a < b && b <= 0                                   = - z          -- fourth sextant: -π < Arg(η) < -2π/3 or Arg(η) = π
+        | b <= a && a < 0                                   = ω * z        -- fifth sextant: -2π/3 ≤ Arg(η) < -π/3
+        | otherwise                                         = (1 + ω) * z  -- sixth sextant: -π/3 ≤ Arg(η) < 0
+    negate (a :+ b) = (-a) :+ (-b)
+    fromInteger n = n :+ 0
+    signum z@(a :+ b)
+        | a == 0 && b == 0 = z               -- hole at origin
+        | otherwise        = z `divE` abs z
```
Bodigrim: It would be nice to avoid division here (and for Gaussian integers as well, generally speaking). To do not repeat a long pattern-matching we can have a single function `absSignum :: EisensteinInteger -> (EisensteinInteger, EisensteinInteger)` and `abs = fst . absSignum`, `signum = snd . absSignum`.

Math/NumberTheory/EisensteinIntegers.hs
```diff
@@ -0,0 +1,345 @@
...
+-- | List of all Eisenstein units, counterclockwise across all sextants,
+-- starting with @1@.
+ids :: [EisensteinInteger]
+ids = take 6 (iterate ((1 + ω) *) 1)
+
+-- | Produce a list of an @EisensteinInteger@'s associates.
+associates :: EisensteinInteger -> [EisensteinInteger]
+associates e = map (e *) ids
+
+-- | Takes an Eisenstein prime whose norm is of the form @3k + 1@ with @k@
```
Bodigrim: Is it safe to say that its norm is of form `6k+1`? It would be more specific.

Bodigrim: It is not clear from the comment, what does this function return?

Math/NumberTheory/EisensteinIntegers.hs
```diff
@@ -0,0 +1,345 @@
...
+-- | Conjugate a Eisenstein integer.
+conjugate :: EisensteinInteger -> EisensteinInteger
+conjugate (a :+ b) = (a - b) :+ (-b)
+
+-- | The square of the magnitude of a Eisenstein integer.
+norm :: EisensteinInteger -> Integer
+norm (a :+ b) = a*a - a * b + b*b
+
+-- | Checks if a given @EisensteinInteger@ is prime. @EisensteinInteger@s
+-- whose norm is a prime congruent to @0@ or @1@ modulo 3 are prime.
```
Bodigrim: I suggest to write about norms modulo 6, because it is more specific.

Math/NumberTheory/EisensteinIntegers.hs
```diff
@@ -0,0 +1,345 @@
...
+-- @z+3k = sqrtMod(9k^2-1)@
+-- @z = sqrtMod(9k^2-1) - 3k@
+-- 
+-- * For example, let @p = 7@, then @k = 1@. Square root of @9*1^2-1 modulo 7@ is @1@.
+-- * And @z = 1 - 3*1 = -2 ≡ 5 (mod 7)@.
+-- * Truly, @norm (5 :+ 1) = 25 - 5 + 1 = 21 ≡ 0 (mod 7)@.
+findPrime :: Integer -> EisensteinInteger
+findPrime p = case Moduli.sqrtModMaybe (9*k*k - 1) (FieldCharacteristic (PrimeNat . integerToNatural $ p) 1) of
+    Nothing      -> error "findPrime: argument must be prime p = 6k + 1"
+    Just sqrtMod -> gcdE (p :+ 0) ((sqrtMod - 3 * k) :+ 1)
```
Bodigrim: Can we use the same approach as for Gaussian integers here? Complex GCD is very expensive.

benchmark/Math/NumberTheory/GaussianIntegersBench.hs
```diff
@@ -21,6 +21,6 @@ benchTau n = bench (show n) $ nf (\m -> sum [tau (x :+ y) | x <- [1..m], y <- [0
 
 benchSuite :: Benchmark
 benchSuite = bgroup "Gaussian"
-  [ bgroup "findPrime" $ map benchFindPrime [1000033, 10000121, 100000037, 1000000009, 10000000033, 100000000057, 1000000000061, 10000000000037]
+  [ bgroup "findPrime" $ map benchFindPrime [1000033, 10000121, 100000037, 1000000021, 10000000033, 100000000057, 1000000000061, 10000000000037]
```
Bodigrim: What was the reason for change?

Bodigrim:
Overall it looks superb incredibly cool. Thanks for good work.

benchmark/Math/NumberTheory/GaussianIntegersBench.hs
rockbmb: Typo. I was using this file by mistake to write newer primes with the same order of magnitude but of the form `3k + 1` which Eisenstein integers require, instead of `4k + 1`, which is what Gaussian integers require.

Math/NumberTheory/EisensteinIntegers.hs
rockbmb: I forgot to leave a not to myself, but using the same process as Gaussian integers makes the process fail. Regarding the complexity of `gcdE`, I found [this](http://cs.au.dk/~gudmund/Documents/cubicres.pdf) link. Worth considering.

Math/NumberTheory/EisensteinIntegers.hs
rockbmb: I don't think it is necessary. Counting with associates, there are 6 primes with norm 3, and 6 with norm 4. These norms are not congruent to 0 or 1 modulo 6.

Math/NumberTheory/EisensteinIntegers.hs
rockbmb: As below, it is not safe to say its norm is of the form `6k + 1` because of `1 - ω` and `2`.

Math/NumberTheory/EisensteinIntegers.hs
rockbmb: @Bodigrim by the way, would you mind reading the paper I linked here and telling me how Lemma 1 can be practically implemented as an approximate norm function, please?

I did not understand the process behind it, only that it is possible and crucial for a fast Eisenstein `gcd`.

rockbmb: Nevermind, I understood what was needed and the latest commit implements it.

rockbmb: I didn't benchmark it directly, but benchmarks for `findPrime` have gotten considerably slower, and the only thing used by that function which changed was `gcdE`. This change might not be worth it after all.

I tried using the [same method](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/GaussianIntegers.hs#L178) that `Math.NumberTheory.GaussianIntegers.findPrime` uses, and as before, it does not work.

Math/NumberTheory/EisensteinIntegers.hs
```diff
@@ -0,0 +1,421 @@
...
+        | a == 0     = s ++ b'
+        | otherwise  = show a ++ op ++ b'
+        where
+            b' = if abs b == 1 then "ω" else show (abs b) ++ "*ω"
+            op = if b > 0 then "+" else "-"
+            s  = if b > 0 then "" else "-"
+
+instance Num EisensteinInteger where
+    (+) (a :+ b) (c :+ d) = (a + c) :+ (b + d)
+    (*) (a :+ b) (c :+ d) = (a * c - b * d) :+ (b * c + a * d - b * d)
```
Bodigrim: It is worth to save one multiplication: `b * (c - d) + a * d` or `b * c + (a - b) * d`.

Math/NumberTheory/EisensteinIntegers.hs
```diff
@@ -0,0 +1,421 @@
...
+-- | Remove @1 - ω@ factors from an @EisensteinInteger@, and calculate that
+-- prime's multiplicity in the number's factorisation.
+divideByThree :: EisensteinInteger -> (Int, EisensteinInteger)
+divideByThree e = go 0 e
+  where
+    go :: Int -> EisensteinInteger -> (Int, EisensteinInteger)
+    go !n !z | r == 0    = go (n + 1) q
+             | otherwise = (n, abs z)
+      where
+        (q, r) = divModE (z * (2 :+ 1)) 3
```
Bodigrim: Can we use `quotEvenI smth 3` instead of `divModE`? It may be worse to inline and reduce multiplication by `2 :+ 1` to avoid any multiplications: `(a :+ b) * (2 :+ 1) = (a + a - b) :+ (a + b)`.

rockbmb:
@Bodigrim the last set of changes improves things, but is still slower than before. An alternative is to check the numbers composing the Eisenstein integer, and only use the new algorithm for magnitudes larger than `10^10` (arbitrarily chosen).

Bodigrim:
I think it is better to revert to the simple GCD algorithm for now.

rockbmb:
@Bodigrim alright. In any case, I fixed things up in my last commit in case this ever becomes useful. Furthermore, it is also generalizable to Gaussian integers.

rockbmb:
@Bodigrim the PR is good to go on my end.

rockbmb:
@Bodigrim after this is merged, since #131 has already been merged, I'll make a new PR to use `solveQuadratic` both with Eisenstein and Gaussian integers' `findPrime`.

Bodigrim:
>  since #131 has also been merged, I'll make a new PR to use `solveQuadratic` 
> both with Eisenstein and Gaussian integers' `findPrime`

The exposed interface of `solveQuadratic` is quite expensive to use in tight loops, so I suggest to leave `findPrime` as is.

---

Merged, awesome!

PR #124, opened 2018-08-11
Added ordered Gaussian primes

### LINKED ISSUE #122: Sort Gaussian primes by norm
It is preferable for applications (e. g., trial division) to list Gaussian primes sorted by ascending norm. Currently this is not so:
```haskell
> take 10 Math.NumberTheory.GaussianIntegers.primes
[1+ι,3,2+ι,1+2*ι,7,11,3+2*ι,2+3*ι,4+ι,1+4*ι]
> map norm $ take 10 Math.NumberTheory.GaussianIntegers.primes
[2,9,5,5,49,121,13,13,17,17]
```

b-mehta:
Resolves #122. 

Added `orderedPrimes` for ordered Gaussian primes. `primes` may still be faster since `Data.List.partition` does not appear to fuse as a producer, even though it does fuse as a consumer and thus `primes` is retained also. That said, it may be hard to do a side-by-side comparison of the two as they have different use cases.

I've also changed the documentation for `primes`: It does not produce all the Gaussian primes (for instance -1-i is never produced), but it does produce them all up to associates.

### DIFF
--- a/Math/NumberTheory/GaussianIntegers.hs
+++ b/Math/NumberTheory/GaussianIntegers.hs
@@ -37,7 +37,7 @@ module Math.NumberTheory.GaussianIntegers (
 
 import Data.List (mapAccumL, partition)
 import Data.Maybe (fromMaybe)
-import Data.List.Ordered (merge)
+import Data.Ord (comparing)
 import GHC.Generics
 
 import qualified Math.NumberTheory.Moduli as Moduli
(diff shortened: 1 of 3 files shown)


Math/NumberTheory/GaussianIntegers.hs
```diff
@@ -149,6 +151,13 @@ primes = [ g
                          in [x :+ y, y :+ x]
          ]
 
+-- |An infinite list of Gaussian primes. See 'primes'.
+orderedPrimes :: [GaussianInteger]
+orderedPrimes = (1 :+ 1): merge l r
+  where (leftPrimes, rightPrimes) = partition (\p -> p `mod` 4 == 3) (tail Sieve.primes)
+        l = [p :+ 0 | p <- leftPrimes]
+        r = [g | p <- rightPrimes, let x :+ y = findPrime p, g <- [x :+ y, y :+ x]]
```
Bodigrim: Looks like something still goes wrong:

```haskell
> map norm $ take 2 $ drop 75 orderedPrimes
[389,361]
```

Could you please write a test for `orderedPrimes`?

b-mehta: While I'm adding tests, are there any other desirable tests other than checking ordering?

b-mehta:
Sounds good - any other changes you'd suggest?

Bodigrim:
Everything else look good to me.

PR #126, opened 2018-08-13
[#125] Implement Hurwitz zeta function

### LINKED ISSUE #125: Implement Hurwitz Zeta function
After #44 and #120, both the [Riemann zeta function](https://en.wikipedia.org/wiki/Riemann_zeta_function) and the [Dirichlet beta function](https://en.wikipedia.org/wiki/Dirichlet_beta_function) have been implemented.

The [Hurwitz zeta function](https://en.wikipedia.org/wiki/Hurwitz_zeta_function) is another important function that can be used to define both the Riemann zeta and the Dirichlet beta functions, so this issue is about refactoring the two latter functions to use the former.

This should all go into the newly created [`Math.NumberTheory.Zeta`](https://github.com/cartazio/arithmoi/tree/master/Math/NumberTheory/Zeta) namespace.

rockbmb:
Closes #125. Still a WIP.

### DIFF
--- a/Math/NumberTheory/Zeta/Hurwitz.hs
+++ b/Math/NumberTheory/Zeta/Hurwitz.hs
@@ -15,57 +15,90 @@ module Math.NumberTheory.Zeta.Hurwitz
   ) where
 
 import Math.NumberTheory.Recurrencies (bernoulli, factorial)
-import Math.NumberTheory.Zeta.Utils   (skipEvens, suminf)
+import Math.NumberTheory.Zeta.Utils   (skipOdds, suminf)
 
--- | Hurwitz zeta function for integer values.
+-- | Value of Hurwitz zeta function evaluated at @ζ(s, a)@ with
+-- @forall t1 t2 . (Floating t1, Ord t1, Integral t2) => s ∈ t2, a ∈ t1@.
+-- The algorithm used was based on the Euler-Maclaurin formula and was derived
+-- from the <https://dlmf.nist.gov/25.11#iii Digital Library of Mathematical Functions>
+-- by the <https://www.nist.gov/ National Institute of Standards and Technology (NIST)>,
+-- formula 25.11.5.
 zeta :: forall a b . (Floating a, Ord a, Integral b) => a -> b -> a -> [a]
 zeta eps s a = map (\second ->
                        firstSum + constant1 + (negate constant2) + second)
                seconds
   where
+    -- When given @1e-14@ as the @eps@ argument, this'll be
+    -- (length . takeWhile (>= 1) . iterate (/ 10) . recip) 1e-14 == 15@,
+    -- meaning @n@ in formula 25.11.5 will be @15@.
     digitsOfPrecision :: Int
     digitsOfPrecision = length . takeWhile (>= 1) . iterate (/ 10) . recip $ eps
 
+    -- @a + n@
     aPlusN :: a
     aPlusN = a + fromIntegral digitsOfPrecision
 
+    -- @(a + n)^s@
     powOfAPlusN :: a
     powOfAPlusN = aPlusN ^^ s
 
+    -- (a + n) ^ (1 - s)            a + n
+    -- ----------------- = ----------------------
+    --       s - 1          (a + n) ^ s * (s - 1)
     constant1 :: a
     constant1 = aPlusN / (powOfAPlusN * ((fromIntegral s) - 1))
 
+    --        1
+    -- ---------------
+    -- 2 * (a + n) ^ s
     constant2 :: a
     constant2 = recip $ 2 * powOfAPlusN
 
+    --               [      1      ]
+    -- \sum_{k=0}^\n | ----------- |
+    --               [ (a + k) ^ s ]
     firstSum :: a
     firstSum = suminf eps $ map (recip . (^^ s) . (a +) . fromInteger) [0..]
 
+    -- [0!, 1!, 2!, 3! ..]
     factorial' :: [a]
     factorial' = map fromInteger factorial
 
+    -- [ B_2k                ]
+    -- | ----- | k <- [1 ..] |
+    -- [ (2k)!               ]
     evenBernoulliDivByFac :: [a]
     evenBernoulliDivByFac = zipWith
                             (\bern fac -> fromRational $ bern / fromInteger fac)
-                            (skipEvens bernoulli)
-                            (skipEvens factorial)
+                            (tail $ skipOdds bernoulli)
+                            (tail $ skipOdds factorial)
 
+    -- [ [ B_2k                ]  [ B_2k                ]  [ B_2k                ]   ]
+    -- | | ----- | k <- [1 ..] |, | ----- | k <- [2 ..] |, | ----- | k <- [3 ..] | ..|
+    -- [ [ (2k)!               ]  [ (2k)!               ]  [ (2k)!               ]   ]
     evenBernoulliFacFracs :: [[a]]
     evenBernoulliFacFracs = iterate tail evenBernoulliDivByFac
 
+    -- [(s + 2*k - 2)! | k <- [1 ..]]
     secondSumNum :: [a]
-    secondSumNum = skipEvens $ drop (fromIntegral s) factorial'
+    secondSumNum = skipOdds $ drop (fromIntegral s) factorial'
 
+    -- [[(s + 2*k - 2)! | k <- [1 ..]], [(s + 2*k - 2)! | k <- [2 ..]], [(s + 2*k - 2)! | k <- [3 ..]] ..]
     secondSumNums :: [[a]]
     secondSumNums = iterate tail secondSumNum
 
+    -- [(s - 1)! * (a + n) ^ s * (a + n) ^ (2*k - 1) | k <- [1 ..]]
     secondSumDenom :: [a]
     secondSumDenom = map ((factorial' !! (fromEnum s - 1) *))
                          (iterate ((aPlusN * aPlusN) *) (aPlusN * powOfAPlusN))
 
+    -- [[(s - 1)! * (a + n) ^ s * (a + n) ^ (2*k - 1) | k <- [1 ..]], [(s - 1)! * (a + n) ^ s * (a + n) ^ (2*k - 1) | k <- [2 ..]] ..]
     secondSumDenoms :: [[a]]
     secondSumDenoms = iterate tail secondSumDenom
 
+    -- [[ B_2k           (s + 2*k - 2)!           |             ]  [ B_2k           (s + 2*k - 2)!           |             ]   ]
+    -- || ----- --------------------------------- | k <- [1 ..] |, | ----- --------------------------------- | k <- [2 ..] | ..|
+    -- [[ (2k)! (s - 1)! * (a + n) ^ (s + 2*k -1) |             ]  [ (2k)! (s - 1)! * (a + n) ^ (s + 2*k -1) |             ]   ]
     seconds :: [a]
     seconds = zipWith3
               (\l1 l2 l3 ->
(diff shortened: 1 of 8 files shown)


rockbmb:
@Bodigrim is it worth it to implement the Hurwitz zeta function with both real arguments? It'll take some more work as it'll require the Gamma function, which I don't think is implemented.

The Riemann zeta and Dirichlet beta will only need an integral first argument, so for a first step that will be sufficient.

Bodigrim:
Gamma function is available in `gamma` package, but only for `Double` and `Float`. It would be better to have it for any `Floating`. 

I think as a first stab we may keep the first argument integer.

rockbmb:
>but only for Double and Float.
Yikes, that won't do. I'll look into submitting a PR to change that, although I won't know if it'll happen. For now, as you agreed, first integer is `Integral`.

Math/NumberTheory/Zeta/Hurwitz.hs
```diff
@@ -0,0 +1,86 @@
...
+import Math.NumberTheory.Recurrencies (bernoulli, factorial)
+import Math.NumberTheory.Zeta.Utils   (skipOdds, suminf)
+
+-- | Value of Hurwitz zeta function evaluated at @ζ(s, a)@ with
+-- @forall t1 t2 . (Floating t1, Ord t1, Integral t2) => s ∈ t2, a ∈ t1@.
+-- The algorithm used was based on the Euler-Maclaurin formula and was derived
+-- from the <https://dlmf.nist.gov/25.11#iii Digital Library of Mathematical Functions>
+-- by the <https://www.nist.gov/ National Institute of Standards and Technology (NIST)>,
+-- formula 25.11.5.
+zeta :: forall a b . (Floating a, Ord a, Integral b) => a -> b -> a -> a
```
Bodigrim: Let us name it `zetaHurwitz` or something else, different from `zeta`. Just to avoid confusion with Riemann zeta.

Bodigrim:
`zetasProperty2` is tough to satisfy (I remember having extremely hard times with it), but it is the only way to check the precision, which `zetas` was asked for.

rockbmb:
@Bodigrim I think the reason the tests fail is that with [4.8.5](http://fredrikj.net/thesis/thesis.pdf#%5B%7B%22num%22%3A85%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C185.945679%2C273.187042%2Cnull%5D) is because in the ["Evaluating the error bound"](http://fredrikj.net/thesis/thesis.pdf#%5B%7B%22num%22%3A87%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C90.0675%2C595.631653%2Cnull%5D) section, it is stated that

> For a precision of `P` bits, we should choose `N ∼ M ∼ P`

However, the [current method](https://github.com/cartazio/arithmoi/blob/91581066e675651f4145d87fd9d1ec79a76b3e05/Math/NumberTheory/Zeta/Hurwitz.hs#L34) to set `N` and `M` is crude, just relying on the fact that `P` bits of precision will result in `P * (log 2 / log 10)` decimal digits of precision.

I think my next step will be to implement the error bounding described in [Theorem 4.8.1](]http://fredrikj.net/thesis/thesis.pdf#%5B%7B%22num%22%3A793%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C90.0675%2C178.754272%2Cnull%5D), and from this bounding derive `N, M`. Ultimately this method won't be perfect either because we don't have the "ball" arbitrary precision types that the Arb library does.

Bodigrim:
It looks to me that tests fail for `eps` close or smaller then `2^-52`. In such case it is just a numeric noise, coming from `Double` itself, which cannot be cured by computing more terms of a series or whatever. I think it makes sense to restrict `eps >= 2^-50` in these tests.

rockbmb:
@Bodigrim You're right. It's not worth asking for precision greater than `1e-52` on a Double, or `1e-50` for `Fixed Prec50`, for example.

rockbmb:
@Bodigrim now that the Hurwitz zeta function is working for integer arguments, I'm going to try extending it to real arguments, first and then complex. This requires the [`gamma` function](http://fredrikj.net/thesis/thesis.pdf#%5B%7B%22num%22%3A79%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C114.639297%2C598.642944%2Cnull%5D), which has been implemented in [the package](https://hackage.haskell.org/package/gamma) we mentioned before. It seems abandoned, so first I'll just do it locally and then consider a PR.

Bodigrim:
Good! 

I looked at `gamma` package more closely. Good news: it still compiles. Bad news: not only `gamma` itself looks unsupported, but it also depends on two more presumably unsupported packages.

rockbmb:
@Bodigrim there are other issues, such as the relative complexity of the implementation and it's lack of documentation. I'm sure there are a lot of very clever tricks used throughout, but without an explanation it's hard to maintain them.

rockbmb:
@Bodigrim Apologies, I made an incorrect statement: it is **not** necessary to use the gamma function for an extended version of `zetaHurwitz`. The Pochhammer symbol (for the rising factorial) has the same definition for either integer or real values.

However, turning the integer argument into a floating argument reduces precision by default, simple tests like `beta(3)` and `zeta(2)` fail. Not sure why that is, could have to do with numerical stability of Prelude functions like `(**)` or maybe a previous assumption I made in the code that is no longer true... This PR is turning into quite the task.

There's one more matter I'd like to discuss. While the `gamma` package will not be useful to us this time, I'm certain it will be in the future because of its importance in analytical number theory. Do you find it worthwhile to bring the package up to speed, for future utilization? At least an initial PR, to ask the maintainer for his opinion.

Bodigrim:
Yeah, dealing with numerical stability is tough. 

---

With regards to `gamma` package. From my perspective it would be nice to migrate https://github.com/mokus0/gamma/blob/master/extras/LanczosConstants.hs to any existing matrix package (actually, `matrix` will do) and release it as a user-facing module of `gamma`. 

As a zero step, one can add `gamma` and its dependencies to Stackage. 

I typically write to maintainers privately, asking if they have any plans of future development, explaining my intention to contribute and to help with life cycling. People usually like when someone finds their abandoned code useful and is willing to continue development.

rockbmb:
@Bodigrim 
> migrate https://github.com/mokus0/gamma/blob/master/extras/LanczosConstants.hs to any existing matrix package
> and release it as a user-facing module of `gamma`

I'm not sure I understand the idea here. Can you explain what your plan for the `LanczosConstants` module would be?

Bodigrim:
AFAIU from skimming through `gamma`, it uses Lancsoz approximation to compute values of the gamma function. Namely, there are two (one for `Float` and one for `Double`) precomputed lists of coefficients, which allow to compute `gamma` up to the precision of either `Float` or `Double` correspondingly. These lists of coefficients were generated by `LanczosConstants` module. 

If one will be able to run `LanczosConstants` for different floating types, he will be able to generate relevant coefficients and derive implementations of `gamma` for other types. 

As you noted above, we do not need this at the moment for `arithmoi`. And I have not dig much into `gamma`. Just thinking aloud.

Bodigrim:
I meant "migrate `LanczosConstants` [from an unreleased library] to `matrix`". Currently `LanczosConstants` relies on an unreleased library for matrix computations, which has prevented the origin author to upload it on Hackage as a part of `gamma`.

rockbmb:
Oh, I understand - you meant to migrate the matrix dependency the module relies on, not the module itself.

rockbmb:
By the way, I'm awaiting a reply on `gamma`, but for now I'll focus on closing this PR.

It'll be difficult to make the Hurwitz zeta function numerically stable on real inputs, that can be a patch for a different occasion. For now I'll keep the first argument integral. I'll push what I have now regardless for future reference, same as what I did with the Eisenstein integer `gcd` algorithm, and then revert it.

rockbmb:
@Bodigrim I've been reading the section on bounding the error for the Hurwitz zeta function, Arb's documentation, and I think that at this point it is not worth it to implement error bounding anymore complex than what is already there.

An idea I've been considering lately is to create implement ball arithmetic types in GHC (in the Prelude, alongside `Double` and `Float`, perhaps?), and then write a small set of C FFI bindings to the Arb library (or any other that offers the same functionality, although Arb seems very good and, for now, maintained) only for the functions `arithmoi` currently uses. If there's a function `truncate :: Floating a => Arb_t -> a`, where `Arb_t` is a Haskell wrapper over `arb_t` from the Arb library, then the end-user can choose what type to run the computations with.

As such, this PR is good to go on my end.

Bodigrim:
I use `stack haddock`.

> In any case, this PR is good to go on my end.

Good. I'll take a look during the next week.

Bodigrim:
Please rebase atop latest master.

What I am worried about is that benchmarks of zeta/beta functions (kinda "sum first 20 values of Riemann zeta function) became significantly slower. This may be connected not only to possibly worse convergence properties, but also to the fact that `zetasOdd` recycle previous values to compute next ones, saving time. I believe it may be worth to leave implementation of Riemann zeta function as is, and add a test checking that it matches Hurwitz zeta function.

rockbmb:
@Bodigrim that's probably because I implemented `zetaHurwitz` as a function and not as a recurrence, such as what `zetas` is now. I'll see if I can turn it into one, if not I'll implement your suggestion.

rockbmb:
@Bodigrim I have returned to fix the issue with `zetaHurwitz`. The precision seems to have been kept, and now that there's more reusing of previous values as was done in `zetasOdd/betasEven`, performance should not be dissimilar (hopefully in the same order of magnitude).

I'll try benchmarking it now, but I'll be slightly skeptical about the numbers (running WSL).

rockbmb:
Went from 10x slower to 2x, I guess that's barely acceptable. There might be room for improvement, @Bodigrim can you review the PR once more?

Bodigrim:
It is certainly acceptable, good job! Overall it looks good to me, but I am wondering: do you know what exactly improved performance 5x? My expectation was that Hurwitz zeta functions on consequent integers do not share much, so I am quite surprised.

rockbmb:
@Bodigrim Here's a short explanation:
* From the 4.8.5 formula, we see that the only variables that vary as a result of the arguments passed to the Hurwitz zeta function are `s, a, N, M`. Since `N ~ M` and the algorithm I wrote is quite simple, using no clever ideas to find an appropriate `N` and just assuming `N = M`, what happens is that `s, a, N` are what we care about.
* For a particular kind of function, `a` will not change: if we're calculating a sequence of values of the Riemann zeta function at odd integers, `a` will be `1`. If we're calculating a sequence of values of the Dirichlet beta function at even integers, `a` will be both `0.25` and `0.75` (this is why the benchmarks for Dirichlet beta are 2x as slow as Riemann's - it has to calculate two recurrences, one for `a = 0.25` and one for `a = 0.75`).
* For the same desired precision, `N` will not change. If we pass `1e-15` to `zetas` or `betas`, then `N`'s value will be the same for `s=0, 1, 2, 3 ..`.

This means the only variable that iterates is `s`, giving the opportunity for a lot of sharing of previous values. For example, the fraction involving even indices of Bernoulli's recurrence divided by factorials of even numbers was being recalculated every time `zetaHurwitz` was called. The same goes for rising factorials, and every term involving `(a + N)`. Every such term would be calculated several times for each call of `zetaHurwitz`, but now it only happens once in the body of the `zetaHurwitz` recurrence.

rockbmb:
I should also add that although there is a performance penalty, there is a precision increase (see 4804b57)

rockbmb:
Remembered to fix Haddock, unlike with the `EisensteinInteger`s module that I only fixed after the fact.
@Bodigrim PR is good to go (unless you want to review it further).

Bodigrim:
Thanks for awesome work!

rockbmb:
@Bodigrim No problem. I should also thank you for taking so much of your time to review my PRs, give advice and merge code :+1: .

Issue #129, opened 2018-08-17
Modular equations


Bodigrim:
Provide high-level functions to solve linear and quadratic equations by modulo. 

Implement `solveLinear` satisfying `propLinear` and `solveQuadratic` satisfying `propQuadratic`.

```haskell
solveLinear :: Mod m -> Mod m -> [Mod m]
solveLinear = undefined 

propLinear :: Mod m -> Mod m -> Bool 
propLinear a b = 
  all 
    (\x -> (a * x + b == 0) == x `elem` solutions) 
    [minBound .. maxBound]
  where 
    solutions = solveLinear a b

solveQuadratic :: Mod m -> Mod m -> Mod m -> [Mod m]
solveQuadratic = undefined 

propQuadratic :: Mod m -> Mod m -> Mod m -> Bool 
propQuadratic a b c = 
  all 
    (\x -> (a * x ^ 2 + b * x + c == 0) == x `elem` solutions) 
    [minBound .. maxBound]
  where 
    solutions = solveQuadratic a b
```

Applications: `solveLinear` can be used instead of `solveCongruence` in #110, `solveQuadratic` may be used to solve `norm (z :+ 1) == 0` in quadratic fields (Gaussian, Eisenstein, etc.) for purposes discussed in #111.

Bodigrim:
I'll assign myself to this issue.

b-mehta:
@Bodigrim For efficient implementation of `solveQuadratic` for prime powers, Hensel's lemma is likely a good approach (and of course for other composites, the Chinese remainder theorem will reduce the problem to this case).

PR #110, opened 2018-06-10
(WIP) Discrete logarithm

Taneb:
This is the work I started at ZuriHac. It implements Pollard's rho algorithm to compute the discrete logarithm. This has the deficiency that there are some solutions it cannot find, such as 2^x == 7 (mod 9), which I would like to resolve before this PR gets merged.

PR #130, opened 2018-08-22
Discrete log

### LINKED ISSUE #88: Discrete logarithms
Once #86 lands into master, it would be nice to implement any non-trivial (or even trivial one!) algorithm for computation of discrete logarithms. I do not know much about this field, but hopefully [wiki](https://en.wikipedia.org/wiki/Discrete_logarithm#Algorithms) could serve a good starting point.

b-mehta:
Discrete logarithms and Dirichlet characters. Resolves #88. In roughly this order, this is my plan for full efficient implementation of both.

- [x] Added `MultMod` type to represent elements of the multiplicative group mod m, with Monoid instance. Added `PrimitiveRoot` type to represent primitive roots of the (cyclic) multiplicative group.
Note these two types together make `discreteLogarithm` a total function. Partial constructors have been added and deconstructors. (I would like feedback on the name `MultMod`, I feel a better name is possible.)
- [x] Amended primitive root code, testing, benching with the new type structures.
- [x] Very simple discrete log code, with test and bench.
- [x] Baby-step giant-step method for discrete logs.
- [x] Additional testing and benching for discrete logs, with prime power cases included also.
- [x] Refactor discrete logs to be efficient on prime powers, using Bach's method described [here](https://math.stackexchange.com/questions/1863037/discrete-logarithm-modulo-powers-of-a-small-prime) and [here](https://www2.eecs.berkeley.edu/Pubs/TechRpts/1984/CSD-84-186.pdf). (Also use the canonical isomorphism for double prime powers).
- [x] Implement Pollard's rho algorithm for discrete logs with prime moduli (#129 would be helpful for this). Notably, Pollard's rho is a Las Vegas algorithm (which is why @Taneb's attempt in #110 did not always return success) so may need to run multiple times - experiment with starting inputs.


Bodigrim:
Cool! I'll take a look by the end of the week.

Math/NumberTheory/Moduli/Class.hs
```diff
@@ -180,6 +191,39 @@ infixr 8 ^%
 -- of type classes in Core.
 -- {-# RULES "^%Mod" forall (x :: KnownNat m => Mod m) p. x ^ p = x ^% p #-}
 
+-- | This type represents elements of the multiplicative group mod m, i.e.
+-- those elemets which are coprime to m. Use @toMultElement@ to construct.
```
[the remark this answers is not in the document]
b-mehta: Thanks for this catch!

Math/NumberTheory/Moduli/Class.hs
```diff
@@ -180,6 +191,39 @@ infixr 8 ^%
...
+  maxBound = MultMod (-1)
+
+-- | Attempt to construct a multiplicative group element.
+isMultElement :: KnownNat m => Mod m -> Maybe (MultMod m)
+isMultElement a = if getVal a `gcd` getMod a == 1
+                     then Just $ MultMod a
+                     else Nothing
+
+-- | Powers can be taken without leaving the multiplicative group.
+powMultMod :: (KnownNat m, Integral a) => MultMod m -> a -> MultMod m
```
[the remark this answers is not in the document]
b-mehta: Sure, sounds reasonable.

Math/NumberTheory/Moduli/Class.hs
```diff
@@ -180,6 +191,39 @@ infixr 8 ^%
...
+powMultMod (MultMod a) k
+  | k >= 0 = MultMod (powMod a k)
+  | otherwise = invertGroup (MultMod (powMod a (-k)))
+
+-- | For elements of the multiplicative group, we can safely perform the inverse
+-- without needing to worry about failure.
+invertGroup :: KnownNat m => MultMod m -> MultMod m
+invertGroup = MultMod . fromJust . invertMod . multElement
```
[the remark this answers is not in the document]
b-mehta: Alright, sounds good, fixed.

Math/NumberTheory/Moduli/DiscreteLogarithm.hs
```diff
@@ -0,0 +1,46 @@
...
+--                          in fromIntegral $ fromJust $ elemIndex b vals
+
+discreteLogarithm a b = let n = prefValue . groupSize . getGroup $ a
+                            a' = unPrimitiveRoot a
+                            m = integerSquareRoot (n - 1) + 1 -- simple way of ceiling . sqrt
+                            babies = fromInteger . getVal . multElement <$> iterate (<> a') mempty
+                            table = M.fromList $ zip babies [0..(m-1)]
+                            bigGiant = powMultMod a' (- fromIntegral m)
+                            giants = fromInteger . getVal . multElement <$> iterate (<> bigGiant) b
+                         in head [i*m + j | (v,i) <- zip giants [0..(m-1)], j <- maybeToList $ M.lookup v table]
```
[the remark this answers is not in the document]
b-mehta: My hope is to replace this with a more space-efficient implementation with Pollard's rho, so hopefully this function should not be in the merge and so if it's okay I'll leave this as is for now.
In the future I'll make sure to use `where` clauses in similar scenarios though!

b-mehta:
I'm deeply confused by this failure, any clues on how to fix?

Bodigrim:
> Implement Pollard's rho algorithm for discrete logs with prime moduli (#129 would be helpful for this)

I pushed `solveLinear` into `master`.

b-mehta:
Great, thanks!

Math/NumberTheory/Moduli/Class.hs
```diff
@@ -180,6 +187,39 @@ infixr 8 ^%
 -- of type classes in Core.
 -- {-# RULES "^%Mod" forall (x :: KnownNat m => Mod m) p. x ^ p = x ^% p #-}
 
+-- | This type represents elements of the multiplicative group mod m, i.e.
+-- those elements which are coprime to m. Use @toMultElement@ to construct.
+newtype MultMod m = MultMod { multElement :: Mod m }
```
Bodigrim: I've been never good with names :) 
Maybe `MultGroupMod`?

Math/NumberTheory/Moduli/Class.hs
```diff
@@ -180,6 +187,39 @@ infixr 8 ^%
...
+  mempty = MultMod 1
+  mappend = (<>)
+
+instance KnownNat m => Bounded (MultMod m) where
+  minBound = MultMod 1
+  maxBound = MultMod (-1)
+
+-- | Attempt to construct a multiplicative group element.
+isMultElement :: KnownNat m => Mod m -> Maybe (MultMod m)
+isMultElement a = if getVal a `gcd` getMod a == 1
```
Bodigrim: It is better to use `getNatVal` and `getNatMod` respectively.

Math/NumberTheory/Moduli/DiscreteLogarithm.hs
```diff
@@ -0,0 +1,41 @@
...
+-- | Computes the discrete logarithm. Currently uses a naive search.
+discreteLogarithm
+  :: KnownNat m
+  => PrimitiveRoot m
+  -> MultMod m
+  -> Natural
+-- discreteLogarithm a b = let n = prefValue . groupSize . getGroup $ a
+--                             a' = unPrimitiveRoot a
+--                             vals = genericTake n $ iterate (<> a') mempty
+--                          in fromIntegral $ fromJust $ elemIndex b vals
```
Bodigrim: Remove commented lines?

Math/NumberTheory/Moduli/DiscreteLogarithm.hs
```diff
@@ -0,0 +1,41 @@
...
+-- Maintainer:   Andrew Lelechenko <andrew.lelechenko@gmail.com>
+-- Stability:    Provisional
+-- Portability:  Non-portable
+--
+
+module Math.NumberTheory.Moduli.DiscreteLogarithm where
+
+import Data.Semigroup
+import Data.Maybe
+-- import Data.List
```
Bodigrim: Remove commented lines?

Math/NumberTheory/Moduli/PrimitiveRoot.hs
```diff
@@ -112,6 +119,13 @@ cyclicGroupToModulo = fromFactors . \case
   CGOddPrimePower p k       -> Coprimes.singleton (unPrime p) k
   CGDoubleOddPrimePower p k -> Coprimes.singleton 2 1 <> Coprimes.singleton (unPrime p) k
 
+-- | 'PrimitiveRoot m' is a type which is only inhabited by primitive roots of n.
+data PrimitiveRoot m =
+  PrimitiveRoot { unPrimitiveRoot :: MultMod m -- ^ Extract primitive root value.
+                , getGroup        :: CyclicGroup Natural -- ^ Get cyclic group structure.
+                }
```
Bodigrim: My personal preference is to put constructor on the same line with `data` and list fields indented by two spaces.

benchmark/Math/NumberTheory/DiscreteLogarithmBench.hs
```diff
@@ -0,0 +1,22 @@
+{-# LANGUAGE DataKinds #-}
+module Math.NumberTheory.DiscreteLogarithmBench 
+  ( benchSuite
+  ) where
+
+import Gauge.Main
+import Data.Maybe
+
+-- import Math.NumberTheory.Moduli (Mod, discreteLogarithm, PrimitiveRoot, isPrimitiveRoot, MultMod, isMultElement)
```
Bodigrim: Remove commented lines?

test-suite/Math/NumberTheory/Moduli/DiscreteLogarithmTests.hs
```diff
@@ -0,0 +1,52 @@
...
+import Math.NumberTheory.TestUtils
+
+-- | Ensure 'discreteLogarithm' returns in the appropriate range.
+discreteLogRange :: Positive Natural -> Integer -> Integer -> Bool
+discreteLogRange (Positive m) a b =
+  case someNatVal m of
+    SomeNat (_ :: Proxy m) -> fromMaybe True $ do
+      a' <- isPrimitiveRoot (fromInteger a :: Mod m)
+      b' <- isMultElement (fromInteger b)
+      return $ discreteLogarithm a' b' < totient m
```
Bodigrim: It is worth to check that it is also `>= 0`.

test-suite/Math/NumberTheory/Moduli/PrimitiveRootTests.hs
```diff
@@ -73,34 +73,34 @@ isPrimitiveRootProperty1 :: AnySign Integer -> Positive Natural -> Bool
...
-               || not (isPrimitiveRoot n')
+               || isNothing (isPrimitiveRoot n')
     InfMod{}   -> False
 
 isPrimitiveRootProperty2 :: Positive Natural -> Bool
 isPrimitiveRootProperty2 (Positive m)
   = isNothing (cyclicGroupFromModulo m)
   || case someNatVal m of
-    SomeNat (_ :: Proxy t) -> any isPrimitiveRoot [(minBound :: Mod t) .. maxBound]
+    SomeNat (_ :: Proxy t) -> not $ null $ mapMaybe isPrimitiveRoot [(minBound :: Mod t) .. maxBound]
```
Bodigrim: I'd rather write `any (isJust . isPrimitiveRoot)`.

Math/NumberTheory/Moduli/Class.hs
b-mehta: Sure, but out of interest why are these preferable?

test-suite/Math/NumberTheory/Moduli/DiscreteLogarithmTests.hs
b-mehta: The return type of `discreteLogarithm` is `Natural`, so it cannot be negative.

Math/NumberTheory/Moduli/Class.hs
Bodigrim: The internal representation of `Mod m` uses `Natural` both for value and for modulo.

test-suite/Math/NumberTheory/Moduli/DiscreteLogarithmTests.hs
Bodigrim: OK, then it is fine.

b-mehta:
The Bach reduction gave a large speedup for prime powers: the case `2^n = 7 mod 3^20` went from around 41ms to 3.5µs and `3^n = 17 mod 5^16` went from 549ms to 3.8µs.  Range benchmarks did not change significantly, which is expected since prime powers are rare compared to primes in general.

Bodigrim:
Cool!

b-mehta:
I've added Pollard's rho but on benchmarking, it doesn't have the speedup I expected. Specifically, for moduli below around 10^8 it is slower than the baby-step giant-step method, but works faster for large moduli. For cases near 10^6, the BSGS method took 27.27ms while Pollard's rho took 43.15ms while for cases near 10^10, BSGS took 8.61s and Pollard's rho took 4.06s. 

While it's not surprising that Pollard's rho is eventually faster, I'm surprised the cutoff is so high. Is there some obvious deficiency in my rho algorithm that makes it perform badly on small inputs? @Bodigrim

(Note the rho implementation isn't perfect right now - it currently only works _probabilistically_ but the probability of error is so small that it succeeds on all the test cases and benchmark cases. The functions `starter` and `begin` are part of the fix for this, but right now they appear redundant, and when failure occurs, the error will be from calling `head` on an empty list in line 79. I'll fix in later commits.)

Bodigrim:
There are two sources of inefficiency: redundant divisions in `step` and allocating lazy triples on heap. 

Here is my attempt:

```haskell
data Pollard = Pollard !Integer !Integer !Integer

discreteLogarithmPrime :: Integer -> Integer -> Integer -> Natural
discreteLogarithmPrime p a b = fromInteger $ head $ filter check $ begin (starter 0 0)
  where
    n = p-1 -- order of the cyclic group

    halfN = n `quot` 2
    mul2 m = if m < halfN then m `shiftL` 1 else m `shiftL` 1 - n

    step (Pollard xi ai bi) = case xi `rem` 3 of
                          0 -> Pollard (xi*xi `rem` p) (mul2 ai) (mul2 bi)
                          1 -> Pollard (a*xi `rem` p)  (ai+1) bi
                          _ -> Pollard (b*xi `rem` p)  ai (bi+1)
                          
    starter x y = Pollard (powModInteger a x n * powModInteger b y n `rem` n) x y
    begin t = go (step t) (step (step t))

    go :: Pollard -> Pollard -> [Integer]
    go tort@(Pollard xi ai bi) hare@(Pollard x2i a2i b2i)
      | xi == x2i = solveLinear' n (bi - b2i) (ai - a2i)
      | otherwise = go (step tort) (step (step hare))

    check t = powModInteger a t p == b
```

One can check by dumping GHC Core `{-# OPTIONS_GHC -ddump-simpl -dsuppress-all -dno-suppress-type-signatures #-}` that no `Pollard` thunk is actually allocated and all computations have reduced to naked `Integer`. Your version of `discreteLogarithmPrime` tended to allocate and deallocate `(_, _, _)` constructors here and there. 

Before:
```
Discrete logarithm/range/100 cases near 10000 mean 15.01 ms  ( +- 2.861 ms  )
Discrete logarithm/range/100 cases near 1000000 mean 124.5 ms  ( +- 26.73 ms  )
Discrete logarithm/range/100 cases near 100000000 mean 976.0 ms  ( +- 87.84 ms  )
```

Now:
```
Discrete logarithm/range/100 cases near 10000 mean 9.049 ms  ( +- 1.375 ms  )
Discrete logarithm/range/100 cases near 1000000 mean 78.45 ms  ( +- 3.424 ms  )
Discrete logarithm/range/100 cases near 100000000 mean 746.1 ms  ( +- 33.00 ms  )
```

b-mehta:
I'm not convinced that allocation of tuples is an issue here, since the Core output shows that unboxed tuples are being used, which has no heap overhead, but I'll investigate further.

Your code doesn't succeed when I test it - there are many missing `rem` calls which cause it to fail. I suspect the improvement in the range benchmarks is due to these calls being omitted when they are truly necessary.

Bodigrim:
Sorry, it is my fault; maybe wrong compilation parameters. I can confirm: your code does not allocate triples as well.

> Your code doesn't succeed when I test it - there are many missing rem calls which cause it to fail.

What do you mean? It seems passing tests in my environment.

b-mehta:
> What do you mean? It seems passing tests in my environment.

Hm this is curious, the first four benchmarks all fail with head of an empty list with your code, but succeed with mine.

Bodigrim:
Weird. This is what I see:
```
Discrete logarithm/individual case/5ⁿ == 8 mod 10^9 + 7 mean 42.95 ms  ( +- 1.763 ms  )
Discrete logarithm/individual case/2ⁿ == 7 mod 3^500 mean 231.4 μs  ( +- 25.78 μs  )
Discrete logarithm/individual case/2ⁿ == 3 mod 10^11 + 3 mean 176.6 ms  ( +- 7.414 ms  )
Discrete logarithm/individual case/3ⁿ == 17 mod 5^350 mean 288.4 μs  ( +- 35.09 μs  )
```

b-mehta:
Very strange, on running `stack bench --ba '--match prefix Disc -q'` I get this error:

```
arithmoi-0.7.0.0: benchmarks
Running 1 benchmarks...
Benchmark criterion: RUNNING...
criterion: Prelude.head: empty list
benchmarking Discrete logarithm/individual case/5ⁿ == 8 mod 10^9 + 7 ... Benchmark criterion: ERROR

--  While building custom Setup.hs for package arithmoi-0.7.0.0 using:
      /Users/bmehta/.stack/setup-exe-cache/x86_64-osx/Cabal-simple_mPHDZzAJ_2.0.1.0_ghc-8.2.2 --builddir=.stack-work/dist/x86_64-osx/Cabal-2.0.1.0 bench criterion "--benchmark-options=--match prefix Disc -q"
    Process exited with code: ExitFailure 1
```

Bodigrim:
I pushed my changes onto CI: it fails with GHC 8.2 and earlier, but succeedes with GHC 8.4 and newer. Very odd. Unfortunately, I am leaving for vacation and won't be able to investigate until the end of next week.

b-mehta:
No worries - thanks for your help regardless.

Bodigrim:
Got it. Prior to GHC 8.4 both `powModInteger` and `recipModInteger` were broken for a negative first argument (https://ghc.haskell.org/trac/ghc/ticket/14085).

That is why after my modification `solveLinear'` started to return weird results in GHC 8.2 and led to an empty `filter check $ begin (starter 0 0)`. I have pushed fixes for `solveLinear'` into `master` and now my implementation works fine for all supported range of GHCs (https://travis-ci.org/Bodigrim/arithmoi/builds/423235835). 20-30% improvement in benchmarks is observed, same to reported above. I suspect it is purely due to changes in `step` function.

Could you please rebase your branch upon the latest `master`?

b-mehta:
It passes tests but I'm not fully convinced of correctness - you've removed the `rem` after `ai+1` and the like, but the remainder still should be taken after these additions, since they are likely to get larger than the modulus. In addition, the expression for `mul2` assumes that the input `m` is reduced modulo `n`, but this is not guaranteed since the `rem` is no longer taken after `ai+1`.

b-mehta:
I've changed some things, removing some of the `rem` calls. When I benchmark, using `m * 2` instead of `shiftL` is noticeably faster, so I've kept it.  Since `Pollard` reduces to identical code as the tuples, I've left that as is also.  

The calls to `mod` in line 92 may not be strictly necessary - I originally kept them because `solveLinear'` was not exported and so I wanted to make sure the input was as expected. Also, it is likely that the `mod` call will reduce arguments into a smaller range, so the function may run faster, but I'm happy to remove them if necessary.

Bodigrim:
`solveLinear'` should work fine for unreduced arguments. Even without reduction by modulo `bi-b2i` and `ai-b2i` are O(n) with probability close to 1, so there should not be any performance hit. I suggest to remove `mod` from line 92 for the sake of clarity.

Bodigrim:
Awesome. I greatly appreciate your work.

Does it make sense to split this PR into two? The first one about discrete log (which is ready to merge, up to my understanding) and the second (future) one about Dirichlet characters. What is your opinion?

b-mehta:
I think this is a good idea - in fact I'd like to split it into three PRs: The currently implemented work on discrete logs (which is ready to merge), possible speed ups on discrete logs (including Pollig-Hellman, which I'm investigating at the minute) and Dirichlet characters.

Something worth pointing out is that the algorithm is probabilistic, as I mentioned in the most recent commit, and so has a (negligible) probability of failing. Currently three starting inputs are used, which cubes the failure probability making it even more negligible, but still non-zero. Possible fixes are to allow all starting inputs on failure, so that the function is guaranteed not to fail, or just to use a small amount and hope for the best (cf line 87 in the last commit). 
What is your opinion on this?

Bodigrim:
It would be nice to find via brute force at least one example, when a) implementation with only one seed fails to find an answer, b) implementation with two seeds fails, c) current implementation with three seeds fails. And add them to tests. 

Can we resort to BSGS, when Pollard rho has failed with three seeds?

b-mehta:
That would be nice, but I fear that the probability of failure is genuinely tiny - testing all primes less than 7000 represents over 5 billion instances of the discrete log problem, and no failure has yet occured.

Possibly, but BSGS uses IntMaps internally and right now BSGS is used anyway when the inputs are less than 10^8 so in that scenario, the values may not fit within an Int. In addition, if Maps are used instead, it is possible that massive maps are made and create memory issues instead.  I'd be more comfortable allowing more seeds instead.

Bodigrim:
I may be wrong, but my impression is that your implementation of Pollard is **not** probabilistic. AFAIU we say that Pollard algorithm _failed_ for a given seed when the algorithm found a collision with `bi == b2i`, for which `solveLinear'` returns a full, unrestricted range. But in our case we just resort to brute force `check`-ing.

b-mehta:
If `bi == b2i`, then `solveLinear'` attempts to solve `0x + a = 0`, and returns the empty list, so we immediately fail.

Math/NumberTheory/Moduli/DiscreteLogarithm.hs
```diff
@@ -0,0 +1,104 @@
...
+discreteLogarithmPrimePollard :: Integer -> Integer -> Integer -> Natural
+discreteLogarithmPrimePollard p a b =
+  case concatMap runPollard [(0,0),(0,1),(1,1)] of
+    (t:_)  -> fromInteger t
+    []     -> error ("discreteLogarithm: pollard's rho failed, please report this as a bug. inputs " ++ show [p,a,b])
+  where
+    n                 = p-1 -- order of the cyclic group
+    halfN             = n `quot` 2
+    mul2 m            = if m < halfN then m * 2 else m * 2 - n
+    step (xi,!ai,!bi) = case xi `rem` 3 of
```
Bodigrim: Each `step` performs two divisions, which I expect to take a lot of time. I wonder whether we can avoid the first one. The algorithm is expected be robust to a somewhat unequal split into three subsets. 

* We can cast `xi` to `Word` and apply `rem` afterwards. At the very least we will avoid horribly expensive division of long `Integer` and in the best scenario compiler will apply [this trick](https://stackoverflow.com/questions/44212301/efficient-mod-3-in-x86-assembly).

* More aggressive approach is to skew distribution even further, looking at only 3 or 4 bits of `xi`. Something like this:

```haskell
case fromInteger xi .&. (7 :: Word) of 
  0 -> (xi*xi `rem` p, mul2 ai, mul2 bi)
  1 -> (xi*xi `rem` p, mul2 ai, mul2 bi)
  2 -> ( a*xi `rem` p,    ai+1,      bi)
  3 -> ( a*xi `rem` p,    ai+1,      bi)
  4 -> ( a*xi `rem` p,    ai+1,      bi)
  _ -> ( b*xi `rem` p,      ai,    bi+1)

b-mehta: I like this idea! I recall reading that the expected time taken for Pollard can be improved to the expected random bound by using more partitions - I'm sure I read that tighter bounds were achieved with 20 but I can't seem to find that result now. With that in mind, it may be interesting to do something like `fromInteger xi .&. (15 :: Word)`, and have 16 different cases. 

In either case, would it be acceptable to take optimisation work like this into a different PR, and merge (something like) this one for now?

b-mehta: Something like f_T and f_C from (Speeding up pollard's rho method for discrete logarithms, Teske 1998), or the walks described in 2.2.2 and 2.2.3 [here](https://maths-people.anu.edu.au/~brent/pd/rpb231.pdf).

Bodigrim: > In either case, would it be acceptable to take optimisation work like this into a different PR, and merge (something like) this one for now?

Sure. Could you please add a `-- TODO` comment with your ideas and links? It is easy to lose them in the depth of a closed PR.

Bodigrim:
> Note that Math.NumberTheory.Moduli.Equations exports solveLinear' - hopefully this is acceptable

I've thought about it and decided to revert this new export in 473a5ef25eadc4a0d033a691e4e54d24dae5d5fe. An additional public function needs documentation, tests and is a commitment to support its interface. Being public, `solveLinear'` should deal somehow with negative modulo and zero modulo, it may also be reasonable to make it working for any `Integral a` instead of `Integer` only. That said, from the maintainer's viewpoint it is preferable to use `solveLinear` only.

Bodigrim:
Unchecked preconditions are the root of all evil :) They are the worst plague of `arithmoi` API in its current state. 

Currently I am tying up loose ends for an upcoming release; I'll be glad to return to `solveLinear'` question afterwards. Please raise an issue to facilitate discussion.

b-mehta:
> Unchecked preconditions are the root of all evil :) They are the worst plague of `arithmoi` API in its current state.

I see! Fair enough - I have no real complaints about the state of `solveLinear'` as it is.

PR #132, opened 2018-08-27
depend on transformers rather than mtl

strake:


Bodigrim:
Could you please clarify the motivation of the proposed change?

strake:
All the features of mtl used in arithmoi are merely re-exports from transformers. This PR would simplify the dependency graph of packages depending on arithmoi.

arithmoi.cabal
```diff
@@ -43,8 +43,8 @@ library
     ghc-prim <0.6,
     integer-gmp <1.1,
     integer-logarithms >=1.0,
-    mtl >=2.0 && <2.3,
     random >=1.0 && <1.2,
+    transformers >=0.0.1 && <0.6,
```
Bodigrim: It feels weird to specify such lower bound: `transformers-0.0.1` require `base < 4.3`, while `arithmoi` accepts `base >= 4.7`. I'd prefer to write `transformers >=0.5 && <0.6` to be on the safe side.

Bodigrim:
Thanks for your contribution, merged.

PR #133, opened 2018-08-30
Euclidean Domains

### LINKED ISSUE #128: Euclidean rings
This is to avoid messy `div{E,G}`, `mod{E,G}`, `gcd{E,G}` (cf. the implementation of Gaussian integers and #121) and unify them with `div`, `mod`, `gcd` for usual `Integral` types.

```
class Euclidean a where
  quotRem :: a -> a -> (a, a)
  divMod  :: a -> a -> (a, a)

quot :: Euclidean a => a -> a -> a
quot = ...

rem :: Euclidean a => a -> a -> a
rem = ...

div :: Euclidean a => a -> a -> a
div = ...

mod :: Euclidean a => a -> a -> a
mod = ...

instance Integral a => Euclidean a where 
  ...

instance Euclidean GaussianInteger where
  ...

instance Euclidean EisensteinInteger where
  ...
```

Then we can replace `Integral` constraint with `Euclidean` in `Math.NumberTheory.GCD`, `Math.NumberTheory.Prefactored` and `Math.NumberTheory.SmoothNumbers`, making them both more general and applicable to quadratic fields.

https://en.wikipedia.org/wiki/Euclidean_domain

rockbmb:
Closes #128. Missing:

- [x] Refactor `EisensteinIntegers` module to use `EuclideanDomain`.

- [x] Refactor `GaussianIntegers` module to use `EuclideanDomain`.

~~- [ ] Replace `Integral` with  `Euclidean` in `Math.NumberTheory.GCD`.~~

~~- [ ] Replace `Integral` with  `Euclidean` in `Math.NumberTheory.Prefactored`.~~

~~- [ ] Replace `Integral` with  `Euclidean` in `Math.NumberTheory.SmoothNumbers`.~~

### DIFF
--- a/Math/NumberTheory/EisensteinIntegers.hs
+++ b/Math/NumberTheory/EisensteinIntegers.hs
@@ -23,7 +23,6 @@ module Math.NumberTheory.EisensteinIntegers
   , ids
 
   , divideByThree
-  , gcdE
 
   -- * Primality functions
   , factorise
@@ -37,7 +36,7 @@ import Data.Maybe                                      (fromMaybe)
 import Data.Ord                                        (comparing)
 import GHC.Generics                                    (Generic)
 
-import qualified Math.NumberTheory.Euclidean      as ED
+import qualified Math.NumberTheory.Euclidean            as ED
 import qualified Math.NumberTheory.Moduli               as Moduli
 import Math.NumberTheory.Moduli.Sqrt                    (FieldCharacteristic(..))
 import qualified Math.NumberTheory.Primes.Factorisation as Factorisation
@@ -163,18 +162,6 @@ divideByThree = go 0
         (q1, r1) = divMod (a + a - b) 3
         (q2, r2) = divMod (a + b) 3
 
--- | Compute the GCD of two Eisenstein integers. The result is always
--- in the first sextant.
-gcdE :: EisensteinInteger -> EisensteinInteger -> EisensteinInteger
-gcdE g h = gcdE' (abs g) (abs h)
-
-gcdE' :: EisensteinInteger -> EisensteinInteger -> EisensteinInteger
-gcdE' g h
-    | h == 0    = g -- done recursing
-    | otherwise = gcdE' h (abs (g `mod'` h))
-  where
-    mod' = ED.mod :: EisensteinInteger -> EisensteinInteger -> EisensteinInteger
-
 -- | Find an Eisenstein integer whose norm is the given prime number
 -- in the form @3k + 1@ using a modification of the
 -- <http://www.ams.org/journals/mcom/1972-26-120/S0025-5718-1972-0314745-6/S0025-5718-1972-0314745-6.pdf Hermite-Serret algorithm>.
@@ -196,7 +183,7 @@ gcdE' g h
 findPrime :: Integer -> EisensteinInteger
 findPrime p = case Moduli.sqrtModMaybe (9*k*k - 1) (FieldCharacteristic (PrimeNat . integerToNatural $ p) 1) of
     Nothing      -> error "findPrime: argument must be prime p = 6k + 1"
-    Just sqrtMod -> gcdE (p :+ 0) ((sqrtMod - 3 * k) :+ 1)
+    Just sqrtMod -> ED.gcd (p :+ 0) ((sqrtMod - 3 * k) :+ 1)
     where
         k :: Integer
         k = p `div` 6
--- a/Math/NumberTheory/EisensteinIntegers.hs
+++ b/Math/NumberTheory/EisensteinIntegers.hs
@@ -22,14 +22,6 @@ module Math.NumberTheory.EisensteinIntegers
   , associates
   , ids
 
-  -- * Division and remainder functions
-  , divE
-  , divModE
-  , modE
-  , quotRemE
-  , quotE
-  , remE
-
   , divideByThree
   , gcdE
 
@@ -45,6 +37,7 @@ import Data.Maybe                                      (fromMaybe)
 import Data.Ord                                        (comparing)
 import GHC.Generics                                    (Generic)
 
+import qualified Math.NumberTheory.EuclideanDomain      as ED
 import qualified Math.NumberTheory.Moduli               as Moduli
 import Math.NumberTheory.Moduli.Sqrt                    (FieldCharacteristic(..))
 import qualified Math.NumberTheory.Primes.Factorisation as Factorisation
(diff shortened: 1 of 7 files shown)


rockbmb:
@Bodigrim is it worth it to have `instance Integral a => EuclideanDomain a`? Having to use `FlexibleInstances/UndecibleInstances` is something I'd rather avoid.

Bodigrim:
Let us name the class `Eucledian` instead of `EucledianDomain`. This is more in line with other adjectives `Real`, `Fractional`, `Float`, `Integral`, etc.

I am sure that `UndecidableInstances` are fine in this case. IMHO it is better than defining `Euclidean` for every `{Int,Word}{8,16,32,64}`.

rockbmb:
@Bodigrim can you take a look at the error produces when running `cabal (new-)build` on the latest (`241a204`) commit? I am not sure where this is coming from, neither `EisensteinInteger` nor `GaussianInteger` have an `Integral` instance.

One possible solution is to move `instance \`instance Integral a => Euclidean a\`` into a separate file and only import it when needed but that's a last resort.

Bodigrim:
Add `{-# OVERLAPPABLE #-}` pragma:

```haskell
instance {-# OVERLAPPABLE #-} Integral a => Euclidean a where ...
```

rockbmb:
@Bodigrim a couple of questions:
* Is it not possible to add `gcd` to the definition of `Euclidean`? It's mentioned in #128 alongside `divMod{E,G}, quotRem{E,G}`, so I think it should also be included in the class.
* Is it really necessary to replace `Integral` with `Euclidean` in `Math.NumverTheory.SmoothNumbers`? The use of `toInteger` [here](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/SmoothNumbers.hs#L70) makes it difficult because it cannot be elided, so some functions will still require an `Integral` instance. Having to use both `Integral` and `Euclidean` does not seem worth the bother.

Bodigrim:
* My impression was that a single implementation of `gcd` (basically, copy-pasted from `Prelude`) should work for all `Euclidean` types equally well. I may be mistaken, of course. 
* You are right, `fromSmoothUpperBound` should be left as is. But `fromSet`, `fromList`, `smoothOver` and `isValid ` may be generalised AFAIU.

rockbmb:
* You're correct on that, my implementation of `gcdE` was copied from `gcdG` which was itself inspired by `Prelude.gcd`. My question wasn't if this should be done, but if we should avoid this `gcdE/gcdG` duplication  by adding a `gcd` method to `Euclidean`.
* Alright, I'll take this as a reference for the other modules in case similar issues arise.

Bodigrim:
There might be a misundertanding: I meant to copy-paste `gcd` from `Prelude` to `Math.NumberTheory.GCD`, replacing `Integral` constraint by `Euclidean`. And then use `Math.NumberTheory.GCD.gcd` instead of `Prelude.gcd` everywhere through `arithmoi`.

rockbmb:
@Bodigrim Ah, I understand now. Alright, sounds like a plan too, since it also allows me to delete `gcdE/gcdG` in favor of a single `gcd` function.

rockbmb:
@Bodigrim A few functions in `Math.NumberTheory.GCD` ([`coprime, gcdOdd, oddGCD`](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/GCD.hs#L184) rely on [`Math.NumberTheory.Utils.shiftToOdd`](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/Utils.hs#L114), which uses `fromIntegral`. Thus the `Integral` instantiation seems inevitable, and it in turn makes `Euclidean` redundant in my opinion, unless we want to either add the class to the type signatures anyway, or duplicate the entire module.

Can you take a look at `d2e37d2`? I think said commit should conclude this PR in particular. Adding `Euclidean` to other modules is starting to seem like a non-trivial task because of the amount of `Integral/Euclidean` juggling it requires, as such it should be done piecemeal.

Bodigrim:
I see. Thanks for your analysis. Let us postpone this until next Sunday, when I'll be back from vacation, but there is a huge chance that you are right and we should stop here.

rockbmb:
@Bodigrim I apologize, I did not know you were vacationing. If I did, I wouldn't have bothered you so much. Enjoy your holidays, we can continue this next Sunday.

Bodigrim:
No worries, it’s fine.

Bodigrim:
No worries, it’s fine.

Bodigrim:
I made up my mind. Let us proceed with your early suggestion to bake gcd into Eucledian type class (with a default implementation) and merge it.

rockbmb:
@Bodigrim It is not possible to move `gcd` into `Euclidean` without adding a `(Eq a, Num a) =>` constraint. Is this acceptable?

Bodigrim:
Sure, it is totally reasonable.

rockbmb:
@Bodigrim should be good to go now.

Bodigrim:
Merged, many thanks.

I actually have some ideas how to proceed further, but I need to do more experiments beforehand.

rockbmb:
@Bodigrim will keep a watchful eye.

Bodigrim:
Just pushed 411b5e521726d1cd4b3581c8b27055168eca3373, continuing your work here. 

1. You were right, writing `instance Integral a => Euclidean a` was a bad idea. It breaks type inference totally. I ended up with `WrappedIntegral` newtype, same to [WrappedMonad](http://hackage.haskell.org/package/base-4.11.1.0/docs/Control-Applicative.html#t:WrappedMonad), [WrappedArrow](http://hackage.haskell.org/package/base-4.11.1.0/docs/Control-Applicative.html#t:WrappedArrow) and [WrappedMonoid](http://hackage.haskell.org/package/base-4.11.1.0/docs/Data-Semigroup.html#t:WrappedMonoid).

2. The reason, why I asked you to put `gcd` into the type class, was that there are different optimised implementations of `gcd` for `Int`, `Word` and `Integer`. `GHC.Real` uses [{-# RULES #-}](http://hackage.haskell.org/package/base-4.11.1.0/docs/src/GHC.Real.html#line-644) to dispatch between them, but as long as we have a type class already it is cleaner to use a mechanism of instances.

3. I decided to put `quot`, `rem`, `div`, `mod` inside the type class (mirroring the design of `Integral`), because there are special calls for `Int`, `Word` and `Integer`, which are faster than `(fst .) . quotRem` etc.

4. I added `coprime` to `Eucledian`: you can implement it via `gcd`, but for `Integral` types it makes sense to check oddity of both arguments first. I also added `extendedGCD`: apart from the generic implementation, in future we will be able to use `gcdExtInteger`. Altogether it makes `Eucledian` a next generation of `Math.NumberTheory.GCD`.

5. It appeared difficult to employ `Euclidean` for `Math.NumberTheory.SmoothNumbers`. I made some initial progress, but further work is needed to generate smooth `GaussianIntegers`, for instance.

PR #138, opened 2018-09-21
Extend 'smoothOver' and 'smoothOverInRangeBF' to 'Euclidean' types

### LINKED ISSUE #137: Smooth Gaussian and Eisenstein numbers
We say that the number `n` is `B`-smooth if `n` can be represented as a product of powers of elements of a set `B`. Module `Math.NumberTheory.SmoothNumbers` contains functions to deal with smooth numbers. It currently supports `Integral` types only. It would be nice to extend this notion to any `Eucledian` type, including `GaussianInteger` and `EisensteinInteger`. 

The main issue is that there is no natural order on `GaussianInteger`, so it is impossible for `smoothOver` to return smooth numbers in ascending order. For the same reason we cannot have `smoothOverInRange` or `smoothOverInRangeBF`, because for an arbitrary `Eucledian` there is no natural definition of a range.

I expect that we need define two new functions. 

1. Define 

```
smoothOver' 
  :: (Euclidean a, Ord b)
  => (a -> b)
  -> SmoothBasis a 
  -> [a]
smoothOver' norm basis = undefined
```

which outputs numbers, smooth over `basis`, in an ascending order determined by user-supplied `norm`. Then define `smoothOver = smoothOver' abs`.

2. Define 

```
isSmooth
  :: Euclidean a 
  => SmoothBasis a 
  -> a 
  -> Bool
```

and express `smoothOverInRangeBF` via it.

Reference: https://en.wikipedia.org/wiki/Smooth_number

rockbmb:
Closes #137.


Bodigrim:
Cool! Could you please add tests for Gaussian and Eisenstein integers?

rockbmb:
@Bodigrim by the way this is good to go on my end.

Bodigrim:
Sorry for delay, I'll take a look tomorrow.

Math/NumberTheory/SmoothNumbers.hs
```diff
@@ -76,29 +79,36 @@ fromSmoothUpperBound n = if (n < 2)
...
 
+-- | Generate an infinite ascending list of
+-- <https://en.wikipedia.org/wiki/Smooth_number smooth numbers>
+-- over a given smooth basis.
+--
+-- >>> import Data.Maybe
+-- >>> take 10 (smoothOver (fromJust (fromList [2, 5])))
+-- [1, 2, 4, 5, 8, 10, 16, 20, 25, 32]
+smoothOver :: (E.Euclidean a, Ord a) => SmoothBasis a -> [a]
+smoothOver = smoothOver' abs
```
Bodigrim: The problem with `abs` is that it does not satisfy a prerequisite of `smoothOver'` stated in the comment above. For instance, `abs (2 + ι) > abs ((2 + ι) * (1 + ι))`. It means that `mergeListLists` is fed with unsorted (in terms of `abs`) lists and consequently returns unsorted (again in terms of `abs`) results.

For example, notice an unexpected position of `3+4*ι`:
```haskell
> take 10 $ map abs $ smoothOver $ fromJust $ fromList [1+ι, 2+ι]
[1,1+ι,1+3*ι,2,2+ι,2+2*ι,2+6*ι,3+4*ι,2+11*ι,4]
```

I suggest to leave `smoothOver` constrained by `Integral a`, in which case `abs` is a valid norm. And expose `smoothOver'` to users as a function, which works for any `Euclidean`.

rockbmb:
@Bodigrim should be good to go now.

Math/NumberTheory/SmoothNumbers.hs
```diff
@@ -76,29 +80,42 @@ fromSmoothUpperBound n = if (n < 2)
...
-smoothOver :: Integral a => SmoothBasis a -> [a]
-smoothOver pl = foldr (\p l -> mergeListLists $ iterate (map (p*)) l) [1] (unSmoothBasis pl)
+-- | Helper used by @smoothOver@ (@Integral@ constraint) and @smoothOver'@
+-- (@Euclidean@ constraint) Since the typeclass constraint is just
+-- @Num@, it receives a @norm@ comparison function for the generated smooth
+-- numbers.
+-- This function relies on the fact that for any element of a smooth basis @p@
+-- and any @a@ it is true that @norm (a * p) > norm a@.
+-- This condition is not checked.
+smoothOver'' :: forall a b . (Num a, Ord b) => (a -> b) -> SmoothBasis a -> [a]
```
Bodigrim: Wow! This is existing. Actually, I did not realise that `Euclidean` constraint is redundant here and `Num` is sufficient. Let us stick to `Num` in other function as well.

Math/NumberTheory/SmoothNumbers.hs
```diff
@@ -130,24 +147,33 @@ smoothOverInRange s lo hi
...
 -- | isValid assumes that the list is sorted and unique and then checks if the list is suitable to be a SmoothBasis.
-isValid :: Euclidean a => [a] -> Bool
+isValid :: E.Euclidean a => [a] -> Bool
 isValid pl = length pl /= 0 && v' pl
   where
-    v' :: Euclidean a => [a] -> Bool
+    v' :: E.Euclidean a => [a] -> Bool
     v' []     = True
-    v' (x:xs) = x /= 0 && abs x /= 1 && abs x == x && all (coprime x) xs && v' xs
+    v' (x:xs) = x /= 0 && abs x /= 1 && abs x == x && all (E.coprime x) xs && v' xs
```
Bodigrim: Let us skip the check for coprimality, which is the main reason for `Euclidean` constraint. It is enough to check that there are no duplicates. This will have a price: `go2` above may encounter duplicates, but it isn't too expensive to deal with them.

Math/NumberTheory/SmoothNumbers.hs
```diff
@@ -130,24 +147,33 @@ smoothOverInRange s lo hi
...
+
+-- | @isSmooth@ checks if a given number is smooth under a certain @SmoothBasis@.
+-- Does not check if the @SmoothBasis@ is valid.
+isSmooth :: forall a . E.Euclidean a => SmoothBasis a -> a -> Bool
+isSmooth prs x = mf (unSmoothBasis prs) x
+  where
+    mf :: [a] -> a -> Bool
+    mf _         0 = False
+    mf []        n = n == 1 -- mf means manually factor
```
Bodigrim: Let us relax this check to `abs n == 1`.

Math/NumberTheory/SmoothNumbers.hs
rockbmb: >existing

What do you mean here?

Bodigrim: Sorry, just a silly typo.

Math/NumberTheory/SmoothNumbers.hs
rockbmb: I'm not sure I agree with that. Does it make sense to have `{2,4}`-smooth numbers, or `{2,6}`-smooth?

Math/NumberTheory/SmoothNumbers.hs
rockbmb: @Bodigrim won't dropping the `Integral` constraint in `smoothOver` mean `abs` can no longer be used as norm?

Math/NumberTheory/SmoothNumbers.hs
rockbmb: @Bodigrim good catch. Negative numbers can be smooth.

Math/NumberTheory/SmoothNumbers.hs
rockbmb: Wait, I answered my own question. It does make sense to have `{2,4}/{2,6}`-smooth numbers, it's smoothness over a set. Calling it a basis doesn't make a difference.

Although the comments will need to reflect this change.

Math/NumberTheory/SmoothNumbers.hs
Bodigrim: Let us leave `smoothOver` with `Integral`, but make `smoothOver' == smoothOver''`.

Math/NumberTheory/SmoothNumbers.hs
rockbmb: Then I'll also change comments to reflect this.

Math/NumberTheory/SmoothNumbers.hs
```diff
@@ -76,29 +80,43 @@ fromSmoothUpperBound n = if (n < 2)
...
+        go1 :: [a] -> [a] -> [a]
         go1 (h:t) b = h:(go2 t b)
         go1 _     b = b
 
-        go2 :: Ord a => [a] -> [a] -> [a]
+        go2 :: [a] -> [a] -> [a]
         go2 a@(ah:at) b@(bh:bt)
-          | bh < ah   = bh : (go2 a bt)
+          | norm bh < norm ah   = bh : (go2 a bt)
           | otherwise = ah : (go2 at b) -- no possibility of duplicates
```
Bodigrim: Now it is worth to check for duplicates. Otherwise some smooth numbers may repeat. For example,

```haskell
> take 10 $ smoothOver $ Data.Maybe.fromJust $ fromList [2,4]
[1,2,4,4,8,8,16,16,16,32]
```

(It is also worth to add a test IMHO)

rockbmb: @Bodigrim I really don't like the thought of using `nub ∈ O(N²)` here... I think it's worth a refactor to use a `Set` inside this loop.

rockbmb: Nevermind, I spoke too soon. It was a quick fix in the line below this one, just another guard in the pattern match.

Bodigrim:
Great work! Thanks, merged.

PR #142, opened 2018-09-28
Implement inverseTotient

Bodigrim:
This is still an early draft of a future framework for inversion of multiplicative functions, part of #60, implementing an algorithm along the lines of https://arxiv.org/pdf/1401.6054.pdf 

It is fascinating how powerful polymorphism is. Here is how we can compute a list of inverses of the totient function at 120. These are all numbers `n` such that `totient n == 120`.

```haskell
> inverseTotient 120 :: [Integer]
[155,310,183,366,225,450,175,350,231,462,143,286,244,372,396,308,248]
```

What if there are too many inverses to keep in memory and it is enough only count them? Just switch to another type:

```haskell
> inverseTotient 120 :: Counter Integer
Counter 17
```

Interested in minimal or maximal preimage? No worries.

```haskell
> inverseTotient 120 :: Compose Option Min Integer
Compose (Option (Just (Min 143)))
> inverseTotient 120 :: Compose Option Max Integer
Compose (Option (Just (Max 462)))
```

Two largest preimages? Easy as a pie.

```haskell
> inverseTotient 120 :: Max2 Integer
Max2 (Just 450) (Just 462)
```


Bodigrim:
Actually we have, https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/ArithmeticFunctions/Standard.hs#L112

rockbmb:
@Bodigrim found what I was looking for, nevermind.

Bodigrim:
It is getting tougher than I think. First, there must be a bug, taking into account failing tests for `sigma` function. Second, I need more experiments to achieve a decent performance.

Bodigrim:
OK, it seems to be ready. At least I was able to reproduce all results from the paper, including the inverse sum-of-divisors of 10^1000. @rockbmb @b-mehta and anyone else, please review. 

I expect that there are dark corners, where implementation becomes quite convoluted and obscure for an external observer. Please report such places and I'll add clarifying comments. (Since I am deeply in the context, it is difficult for me to identify them myself.)

rockbmb:
@Bodigrim
> I was able to reproduce all results from the paper inverse sum-of-divisors of 10^1000.
Excellent work. I'll take a look, although I cannot guarantee any interesting insights, this code seems quite profound.

rockbmb:
@Bodigrim I can't reproduce the 10^1000 result, how long did execution take for you? I presume that you didn't do it in ghci, but not even an executable did it for me.

Bodigrim:
It took 4139 sec and 576 Mb or RAM.

b-mehta:
Nice work! I'll take a close look in the next few days.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,354 @@
...
+
+-- | Convert a list of powers of a prime into an atomic Dirichlet series
+-- (Section 4, Step 2).
+atomicSeries
+  :: (Semiring b, Num a, Ord a)
+  => (a -> b)               -- ^ How to inject a number into a semiring
+  -> ArithmeticFunction a c -- ^ Arithmetic function, which we aim to inverse
+  -> PrimePowers a          -- ^ List of powers of a prime
+  -> DirichletSeries c b    -- ^ Atomic Dirichlet series
+atomicSeries point ar (PrimePowers p ks) = case ar of
```
rockbmb: Couldn't you just have used `ArithmeticFunction f g` instead of the `ar` argument and forego the case statement altogether?

rockbmb: I don't think there are any plans to extend or modify this datatype, so pattern matching is unnecessary here IMO.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,354 @@
...
+      :: [PrimePowers a]
+      -> (Prime c, Word)
+      -> ([PrimePowers a], (Maybe (Prime c, Word), [PrimePowers a]))
+    go ts (p, k) = (rs, (Just (p, k), qs))
+      where
+        predicate (PrimePowers q ls) = any (\l -> runTot q l `rem` unPrime p == 0) ls
+        (qs, rs) = partition predicate ts
+
+    runTot :: Prime a -> Word -> c
+    runTot p k = case tot of
```
rockbmb: Same for this line, and there are probably other occurences.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,354 @@
...
+  -- ^ Arithmetic function, which we aim to inverse
+  -> ([(Prime c, Word)] -> [PrimePowers a])
+  -- ^ How to find possible prime factors of the argument
+  -> c
+  -- ^ Value of the arithmetic function, which we aim to inverse
+  -> b
+  -- ^ Preimages
+invertFunction point f invF n
+  = DS.lookup n
+  $ foldl (\ds b -> uncurry processBatch b ds) (DS.fromDistinctAscList []) batches
```
rockbmb: Not sure if it's worth it as an isolated change (meaning it's probably worth doing as part of a performance fixup PR, just fixing this line won't make much of a difference overall), but we should probably make an effort to avoid `foldl` and always rely on `foldl'`.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,354 @@
...
+    factors = factorise n
+    batches = strategy f factors $ invF factors
+
+    processBatch
+      :: Maybe (Prime c, Word)
+      -> [PrimePowers a]
+      -> DirichletSeries c b
+      -> DirichletSeries c b
+    processBatch Nothing xs acc
+      = foldl (DS.timesAndCrop n) acc
```
rockbmb: Same here.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,354 @@
...
+      $ map (atomicSeries point f) xs
+
+    -- This is equivalent to the next, general case, but is faster,
+    -- since it avoids construction of many intermediate series.
+    processBatch (Just (p, 1)) xs acc
+      = DS.filter (\a -> a `rem` unPrime p == 0)
+      $ foldl (DS.timesAndCrop n) acc'
```
rockbmb: Here as well.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,354 @@
...
+      $ map (atomicSeries point f) xs2
+      where
+        (xs1, xs2) = partition (\(PrimePowers _ ts) -> length ts == 1) xs
+        vs = DS.unions $ map (atomicSeries point f) xs1
+        (ys, zs) = DS.partition (\a -> a `rem` unPrime p == 0) acc
+        acc' = ys `DS.union` DS.timesAndCrop n zs vs
+
+    processBatch (Just pk) xs acc
+      = (\(p, k) -> DS.filter (\a -> a `rem` (unPrime p ^ k) == 0)) pk
+      $ foldl (DS.timesAndCrop n) acc
```
rockbmb: Here too.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,354 @@
...
+  => (a -> b)
+  -> a
+  -> b
+inverseTotient point n = invertFunction point totientA invTotient n
+{-# SPECIALISE inverseTotient :: Semiring b => (Int -> b) -> Int -> b #-}
+{-# SPECIALISE inverseTotient :: Semiring b => (Word -> b) -> Word -> b #-}
+{-# SPECIALISE inverseTotient :: Semiring b => (Integer -> b) -> Integer -> b #-}
+{-# SPECIALISE inverseTotient :: Semiring b => (Natural -> b) -> Natural -> b #-}
+
+-- | The inverse 'sigma' 1 function such that
```
rockbmb: Excellent comments.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,354 @@
...
+-- | Wrapper to use in conjunction with 'inverseTotient' and 'inverseSigma'.
+-- Extracts the minimal preimage of function.
+newtype MinWord = MinWord { unMinWord :: Word }
+  deriving (Show)
+
+instance Semiring MinWord where
+  zero = MinWord maxBound
+  one  = MinWord 1
+  plus  (MinWord a) (MinWord b) = MinWord (a `min` b)
+  times (MinWord a) (MinWord b) = MinWord (a * b)
```
rockbmb: TIL this how we define one valid semiring for the `min` operator, I wasn't quite sure `(*)` would work.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
Bodigrim: Good catch!

Bodigrim:
@rockbmb thanks for valuable suggestions.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
rockbmb: I may have missed a few, so for completeness' sake, it's better to use your editor's find and replace feature on this file.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -347,16 +347,17 @@ instance Semiring MaxNatural where
 
 -- | Wrapper to use in conjunction with 'inverseTotient' and 'inverseSigma'.
 -- Extracts the minimal preimage of function.
--- 'Nothing' stands for a positive infinity.
-newtype MinNatural = MinNatural { unMinNatural :: Maybe Natural }
-  deriving (Show)
+data MinNatural
+  = MinNatural { unMinNatural :: !Natural }
```
rockbmb: Right, I didn't give much importance to this before, but why was this `Maybe` before, and why change it now?

b-mehta: It seems to me that switching to a newly defined sum type allows the non-infinite case to be more strict, since the indirection provided by `Just` is removed

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,363 @@
...
+      $ primes
+
+    doPrime :: Prime a -> S.Set Word
+    doPrime (unPrime -> p) = S.fromDistinctAscList
+      [ e
+      | e <- [1 .. intToWord (integerLogBase (toInteger p) (toInteger n))]
+      , n `rem` ((p ^ (e + 1) - 1) `quot` (p - 1)) == 0
+      ]
+
+    pksLarge :: M.Map (Prime a) (S.Set Word)
```
b-mehta: Very minor point: `M.Map` here can be replaced with `Map`.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,363 @@
...
+import Math.NumberTheory.Primes (primes)
+import Math.NumberTheory.UniqueFactorisation
+import Math.NumberTheory.Utils.DirichletSeries (DirichletSeries)
+import qualified Math.NumberTheory.Utils.DirichletSeries as DS
+import Math.NumberTheory.Utils.FromIntegral
+
+data PrimePowers a = PrimePowers
+  { _ppPrime  :: Prime a
+  , _ppPowers :: [Word] -- sorted list
+  }
```
b-mehta: These deconstructors are not used nor exported, are they still necessary? Perhaps a note would suffice to explain the role of both arguments.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,363 @@
...
+{-# SPECIALISE invertFunction :: Semiring b => (Integer -> b) -> ArithmeticFunction Integer Integer -> ([(Prime Integer, Word)] -> [PrimePowers Integer]) -> Integer -> b #-}
+{-# SPECIALISE invertFunction :: Semiring b => (Natural -> b) -> ArithmeticFunction Natural Natural -> ([(Prime Natural, Word)] -> [PrimePowers Natural]) -> Natural -> b #-}
+
+-- | The inverse 'totient' function such that
+--
+-- > all ((== x) . totient) (inverseTotient x)
+-- > x `elem` inverseTotient (totient x)
+--
+-- The return value is parametrized by a semiring, which allows
+-- various applications. E. g., list all preimages:
```
b-mehta: It might be nice to include examples of calculating the sum or sum of squares of preimages:
```
>>> getSum $ inverseTotient Sum 120
4904
>>> getSum $ inverseTotient (Sum . (^2)) 120
1573974
```

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,363 @@
...
+  -- ^ Factorisation of a value of the arithmetic function
+  -> [PrimePowers a]
+  -- ^ Possible prime factors of an argument of the arithmetic function
+  -> [(Maybe (Prime c, Word), [PrimePowers a])]
+  -- ^ Batches, corresponding to each element of the factorisation of a value
+strategy (ArithmeticFunction f g) factors args = (Nothing, ret) : rets
+  where
+    (ret, rets)
+      = mapAccumL go args
+      $ sortBy (\(p1, _) (p2, _) -> p2 `compare` p1) factors
```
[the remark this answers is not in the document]
rockbmb: Nice catch, I hadn't heard of this tip before.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
rockbmb: More minor still: isn't it more consistent to import the `Set` type unqualified to be same as `Map`?

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
Bodigrim: It is a matter of taste, but I prefer to keep record fields named even if deconstructors are not used.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
Bodigrim: Actually `Sum` is redundant: `inverseTotient id 120` works as well.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
b-mehta: Fair enough!

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
b-mehta: True! In that case, it might be nice to add a note saying that the semiring on Word is the one given by addition and multiplication, to explain why that works.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,363 @@
...
+-- >>> import Data.Semigroup
+-- >>> S.map getProduct (inverseTotient (S.singleton . Product) 120) :: S.Set Word
+-- fromList [143,155,175,183,225,231,244,248,286,308,310,350,366,372,396,450,462]
+--
+-- Count preimages:
+--
+-- >>> import Control.Applicative
+-- >>> import Data.Semigroup
+-- >>> inverseTotient (const $ Const 1) 120 :: Const (Sum Word) Word
+-- Const (Sum {getSum = 17})
```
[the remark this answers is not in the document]
Bodigrim: Nice catch!

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
Bodigrim: Done, thanks.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
```diff
@@ -0,0 +1,363 @@
...
+-- > all ((== x) . sigma 1) (inverseSigma x)
+-- > x `elem` inverseSigma (sigma 1 x)
+--
+-- The return value is parametrized by a semiring, which allows
+-- various applications. E. g., list all preimages:
+--
+-- >>> import qualified Data.Set as S
+-- >>> import Data.Semigroup
+-- >>> S.map getProduct (inverseSigma (S.singleton . Product) 120) :: S.Set Word
+-- fromList [54,56,87,95]
```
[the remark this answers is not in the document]
Bodigrim: I am not keen to add more functions, easily expressible in terms of existing ones. It makes API bulky, especially if we add more inverse functions. 

I changed a comment, explicitly mentioning that the embedding must be multiplicative. Hopefully, an alert reader would be able to derive, which semiring to use. Unfortunately, one need to be aware how instances of `Semiring` are defined. It is a bit confusing that `id`, `Sum` and `Product` are the same semiring, but `S.singleton . Sum` and `S.singleton . Product` are different.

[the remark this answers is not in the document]
b-mehta: Yes, explosion of functions is an issue here, but I'd argue that a 'standard' use of this module would be to extract the set of preimages, and the choice of this semiring is nontrivial enough for it to warrant a convenience function. 

How about something higher order:
```
asSet :: (Integral a) => (forall b. Semiring b => (a -> b) -> a -> b) -> a -> S.Set a
asSet f = S.map getProduct . f (S.singleton . Product)
```
then `asSet inverseSigma === preimagesSigma`, and works similarly for `inverseTotient`, and  any other future functions.

Math/NumberTheory/ArithmeticFunctions/Inverse.hs
Bodigrim: We can do even something like `inverseTotient (sigma 1) 120` %)

b-mehta:
For interest, we can also compute the product of preimages:
```
data Prod a = Prod { count :: Word
                   , value :: a
                   }
                   deriving Show

instance Monoid a => Semiring (Prod a) where
  zero = Prod 0 mempty
  one = Prod 1 mempty
  (Prod n a) `plus` (Prod m b) = Prod (n + m) (a <> b)
  (Prod n a) `times` (Prod m b) = Prod (n * m) (stimes m a <> stimes n b)
```
then
```
>>> inverseTotient (Prod 1 . Product) 120             
Prod {count = 17, value = Product {getProduct = 239173504760939546437745014990080000000000}}
```
matches the expected value, and counts the number of preimages as a bonus. (It is not hard to check this forms a commutative semiring, and that the map from `P_fin(Z)` is a weak homomorphism: note that disjointness is required here!). 

Another semiring can be given by wrapping `Set` to ensure it contains no more than a fixed number of elements, to get the first `n` preimages:
```
newtype FirstThree a = Three (S.Set a)
  deriving Show

instance (Ord a, Monoid a) => Semiring (FirstThree a) where
  zero = Three zero
  one = Three one
  (Three a) `plus` (Three b) = Three (S.take 3 (a `plus` b))
  (Three a) `times` (Three b) = Three (S.take 3 (a `times` b))
```
then
```
>>> inverseSigma (Three . S.singleton . Product) 120
Three (fromList [Product {getProduct = 54},Product {getProduct = 56},Product {getProduct = 87}])
```

Bodigrim:
@rockbmb @b-mehta thanks for reviewing!

Bodigrim:
@b-mehta `Prod` and `FirstThree` are very cunning, cool stuff :)

rockbmb:
@b-mehta using your `Data.Ord.Down` tip with the `FirstThree` datatype means we can also find the largest `N` preimages, by doing `newtype LastThree a = Three (S.Set (Down a))`. This is really quite cunning.

PR #145, opened 2018-10-01
[#141] Add integer `n`-freedom functions

### LINKED ISSUE #141: Generate squarefree/cubefree/etc. numbers
A number is called n-free if its factorisation contains no prime powers larger or equal to n. For example, 30 = 2 * 3 * 5 is 2-free (square-free), 3-free (cube-free), etc.; 12 = 2 ^ 2 * 3 and 18 = 2 * 3 ^ 2 are not square-free, but are cube-free; 1000 is not even cube-free, but is 4-free. 

It would be nice to implement two new functions:

1. Predicate `isNFree :: Word -> ArithmeticFunction a Bool` such that `runFunction (isNFree n) x` check whether `x` is n-free.

2. Function `nFrees :: Word -> [a]`, which lists all n-free numbers. I expect that an effective implementation of `nFrees` should avoid division (similar to Eratosthenes sieve).

rockbmb:
Closes #141.

### DIFF
--- a/Math/NumberTheory/ArithmeticFunctions/NFreedom.hs
+++ b/Math/NumberTheory/ArithmeticFunctions/NFreedom.hs
@@ -16,6 +16,7 @@ import Control.Monad.ST                     (runST)
 import qualified Data.Vector.Unboxed         as U
 import qualified Data.Vector.Unboxed.Mutable as MU
 
+import Math.NumberTheory.Powers.Squares     (integerSquareRoot)
 import Math.NumberTheory.Primes             (primes, unPrime)
 import Math.NumberTheory.Utils.FromIntegral (wordToInt)
 
@@ -47,5 +48,8 @@ sieveBlockNFree n lowIndex' len'
     len :: Int
     len = wordToInt len'
 
+    highIndex :: Int
+    highIndex = lowIndex + len - 1
+
     ps :: [Int]
-    ps = take len $ map unPrime primes
+    ps = takeWhile (<= integerSquareRoot highIndex) $ map unPrime primes
--- a/Math/NumberTheory/ArithmeticFunctions/NFreedom.hs
+++ b/Math/NumberTheory/ArithmeticFunctions/NFreedom.hs
@@ -58,6 +58,8 @@ sieveBlockNFree n lowIndex' len'
     ps :: [Int]
     ps = takeWhile (<= integerSquareRoot highIndex) $ map unPrime primes
 
+-- | For a given nonnegative integer power @n@, generate all @n@-free
+-- numbers in ascending order, starting at @1@.
 nFrees :: forall a . (Integral a, UniqueFactorisation a) => Word -> [a]
 nFrees 0 = [1]
 nFrees 1 = [1]
--- a/Math/NumberTheory/ArithmeticFunctions/NFreedom.hs
+++ b/Math/NumberTheory/ArithmeticFunctions/NFreedom.hs
@@ -26,16 +26,23 @@ import Math.NumberTheory.UniqueFactorisation (UniqueFactorisation)
 import Math.NumberTheory.Utils.FromIntegral  (wordToInt)
 
 -- | Evaluate the `isNFreeA` function over a block.
--- Value of @f@ at 0, if zero falls into block, is undefined.
+-- Value at @0@, if zero falls into block, is undefined.
+-- This function should be used with a negative lower bound. If it is, the
+-- result is undefined.
 --
 -- >>> sieveBlockNFree 2 1 10
 -- [True, True, True, False, True, True, True, False, False, True]
 sieveBlockNFree
   :: forall a . Integral a
   => Word
+  -- ^ Power whose @n@-freedom will be checked.
   -> a
+  -- ^ Lower index of the block.
   -> Word
+  -- ^ Length of the block.
   -> U.Vector Bool
+  -- ^ Vector of flags, where @True@ at index @i@ means the @i@-th element of
+  -- the block is @n@-free.
 sieveBlockNFree _ _ 0 = U.empty
 sieveBlockNFree n lowIndex len'
   = runST $ do
@@ -65,22 +72,43 @@ sieveBlockNFree n lowIndex len'
 
 -- | For a given nonnegative integer power @n@, generate all @n@-free
 -- numbers in ascending order, starting at @1@.
+--
+-- When @n@ is @0@ or @1@, the resulting list is @[1]@.
 nFrees :: forall a . (Integral a, UniqueFactorisation a) => Word -> [a]
 nFrees 0 = [1]
 nFrees 1 = [1]
 nFrees n = concatMap nFreesListInternal nFreeList
   where
+    -- The 56th element of @iterate (2 *) 256@ is @2^64 :: Word == 0@, so to
+    -- avoid overflow only the first 55 elements of this list are used.
+    -- After those, since @maxBound :: Int@ is the largest a vector can be,
+    -- this value is just repeated. This means after a few dozen iterations,
+    -- the sieve will stop increasing in size.
     strides :: [Word]
     strides = take 55 (iterate (2 *) 256) ++ repeat (fromIntegral (maxBound :: Int))
+
+    -- Infinite list of lower bounds at which @sieveBlockNFrees@ will be
+    -- applied. This has type @Integral a => a@ instead of @Word@ because
+    -- unlike the sizes of the sieve that eventually stop increasing (see
+    -- above comment), the lower bound at which @sieveBlockNFrees@ does not.
     bounds :: [a]
     bounds = scanl' (+) 1 $ map fromIntegral strides
+
     nFreeList :: [(U.Vector Bool, a, Word)]
     nFreeList =
        zipWith (\lo strd -> (sieveBlockNFree n lo strd, lo, strd)) bounds strides
+
     nFreesListInternal :: (U.Vector Bool, a, Word) -> [a]
     nFreesListInternal (bs, lo, strd) =
-        let strd' :: a
+        let -- When indexing the array of flags @bs@, the index has to be an
+            -- @Int@. As such, it's necessary to cast @strd@ twice.
+            -- Once, immediately below, to create the range of values whose
+            -- @n@-freedom will be tested. Since @nFrees@ has return type
+            -- @[a]@, this cannot be avoided as @strides@ has type @[Word]@.
+            strd' :: a
             strd' = fromIntegral strd
+            -- Twice, immediately below, to create the range of indices with
+            -- which to query @bs@.
             strd'' :: Int
             strd'' = fromIntegral strd
         in map snd .
--- a/Math/NumberTheory/ArithmeticFunctions/NFreedom.hs
+++ b/Math/NumberTheory/ArithmeticFunctions/NFreedom.hs
@@ -41,7 +41,7 @@ sieveBlockNFree n lowIndex' len'
     forM_ ps $ \p -> do
       let pPow = p ^ n
           offset = negate lowIndex `mod` pPow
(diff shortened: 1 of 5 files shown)


rockbmb:
@Bodigrim I can't think of any other way of writing `nFrees` other than `nFrees n = filter (isNFree n) [1..]`.

I thought about what you meant in [this comment](https://github.com/cartazio/arithmoi/issues/141#issuecomment-426053279):

> Basically, one can reuse [runFunctionOverBlock](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/ArithmeticFunctions/SieveBlock.hs#L60) or - for the sake of performance - re-implement it, following the same approach.

But the size of the array has to be know beforehand for `sieveBlock` to work.
One solution to this is return to use `0` when the number is not `n`-free, and then filter the resulting vectors and/or concatenate them into a list.
Another solution is to do `runFunctionOverBlock (isNFreeA n) 1 stride` (where `stride` is how many numbers to process at a time) infinitely, concatenate the resulting vectors and then recurse over the resulting list i.e.

```go !n (bool : bools)
    | bool = n : rec
    | otherwise = rec
  where rec = go (n + 1) bools
```
but this seems a little roundabout. What do you think? Any better ideas?

Bodigrim:
My impression was that listing n-free numbers is very close to sieving primes. Our implementation of Eratosphenes sieve basically generates an infinite list of boolean arrays (`PrimeSieve` structure), then sieves out each 2nd, 3rd, 5th, etc. element, then reports remaining indices in a lazy fashion. For n-free numbers one may follow exactly the same pattern, but remove each (2^n)th, (3^n)th, (5^n)th, etc. element.

rockbmb:
@Bodigrim every function in `Math.NumberTheory.Primes.Sieve.Eratosthenes` is extremely specialized for efficiency. I do not want to modify this code for use in this problem, but I also do not want to reinvent the solution to problem already solved.

I'll probably just write a simpler sieve.

Bodigrim:
For the sake of conistency we may have `sieveBlockNFree  :: Word -> (Word -> Word -> U.Vector Bool)` written along the lines of `sieveBlockMoebius`. And `nFrees`, built atop of `sieveBlockNFree`.

Math/NumberTheory/ArithmeticFunctions/Standard.hs
```diff
@@ -254,6 +257,33 @@ instance Monoid (Mangoldt a) where
...
+-- factorisation into prime factors, no factor has an exponent large than or
+-- equal to @n@.
+isNFreeA :: UniqueFactorisation n => Word -> ArithmeticFunction n Bool
+isNFreeA n = ArithmeticFunction (\_ pow -> All $ pow < n) getAll
+
+nFrees :: forall a . (Integral a, UniqueFactorisation a) => Word -> [a]
+nFrees n = concatMap nFreesListInternal nFreeList
+  where
+    stride :: Word
+    stride = 256
```
Bodigrim: Sieving is efficient only when the length of each individual sieve is at least sqrt of upper bound. We should either choose `stride` large enough, or take sieves of increasing size (256, 512, 1024, etc.).

Math/NumberTheory/ArithmeticFunctions/NFreedom.hs
```diff
@@ -0,0 +1,51 @@
...
+
+  where
+    lowIndex :: Int
+    lowIndex = wordToInt lowIndex'
+
+    len :: Int
+    len = wordToInt len'
+
+    ps :: [Int]
+    ps = take len $ map unPrime primes
```
rockbmb: There's also this incorrect line. I did this because otherwise the function wouldn't terminate, but this'll always choose the same primes and cause `n`-unfree numbers to be incorrectly reported as free if one of their factors is after the 256th prime in `primes.`

This should probably involve `lowIndex` as well.

Math/NumberTheory/ArithmeticFunctions/Standard.hs
rockbmb: I'll address this afterwards.

Math/NumberTheory/ArithmeticFunctions/NFreedom.hs
rockbmb: Fixed.

Math/NumberTheory/ArithmeticFunctions/Standard.hs
Bodigrim: It sounds reasonable to me.

rockbmb:
(Moving discussion here because previous changes are outdates and cannot be directly seen).
@Bodigrim regarding increasing sieve sizes, overflow just occured to me. Doing `iterate (256 *) 256` means at some point, we'll have `0`.

My opinion is that although machine resources are finite and limited, we should aim to return as many valid elements in these infinite sequences.

At this point, there are some problems at hand:
* Sieve sizes **will** overflow, we're using `Word`s. This is acceptable because half of the representation space of `Int`s is useless when trying to represent array sizes, so let's keep it like that. However, the algorithm will still work if sieve sizes overflow, but not to 0, and keep doing so. Performance might take a hit because sieve sizes grow geometrically for some iterations, and then start again from small values. Depending on the initial value chosen or the rate of growth it should be possible to always have nonzero sieve sizes.
* One other problem is that calculating the prime power might overflow again, and cause the wrong elements to be dropped from the sequence. This also has the unintended consequence of the `SIGSEGV`s I mentioned before. I've found a fix for this, which is to do `forM_ (takeWhile (<= highIndex) [offset, offset + pPow .. len - 1])` instead of `forM_ [offset, offset + pPow .. len - 1]` [here](https://github.com/cartazio/arithmoi/pull/145/files#diff-7f9e58f1b39a857ff7432558f30d9d6fR44). This just means that if the prime power is too large to be removed from the current iteration, it'll be removed in the future, when [`offset = negate lowIndex `mod` pPow`](https://github.com/cartazio/arithmoi/pull/145/files#diff-7f9e58f1b39a857ff7432558f30d9d6fR43) is low enough. However, this can have unintended consequences because of what I said above - `len` might not ever be large enough to allow the prime power to be removed.

This last problem can be fixed with a large, but fixed or slow-growing sieve size.

rockbmb:
And there's another problem, which is the impossibility of giving `Integer` type to `lowIndex/highIndex` since they're going to represent the vector's indices. This PR has turned out a little more complicated than I thought.

I'll give it some more thought this weekend.

Bodigrim:
> Doing `iterate (256 *) 256` means at some point, we'll have `0`.

Well, this is a way too fast growing rate. Something like `iterate (* 2) 256` should fit better. 

> And there's another problem, which is the impossibility of giving `Integer` type to `lowIndex`/`highIndex` since they're going to represent the vector's indices.

If you sieve up to `n`, then you have to store in memory a table of primes up to `sqrt n`. That is why I believe that `maxBound :: Word` is a natural applicability bound of sieving. For larger integers one usually do not have enough RAM to store enough primes and must resort to pointwise filtering. 

That said, we may restrict type of `nFrees` to `[Word]` only without losing any potential benefits.

rockbmb:
@Bodigrim Sorry, I made my point incorrectly. It's not just about the rate of sieve sizes growth:

```
let (l, l') = (iterate (2*) (256 :: Word), iterate (256*) (256 :: Word))
let f = length . takeWhile (/= 0)
f l
f l'
```

You should get

```
56
7
```

rockbmb:
I'll progressively improve the function as much as I can, and should it not be much better than now, I'll at least document that e.g. it can only find `n`-free numbers up to a certain `Word`, or that `n` can only be up to a certain power.

rockbmb:
@Bodigrim this is good to go on my end. Managed to solve the issue, and it "only" cost an `Integral` constraint. I'm just going to add a few comments explaining the growth of the sieve and why it stops after a certain point, and then the PR will be done.

Bodigrim:
(Sorry, I am on vacation, will take a look next week)

rockbmb:
@Bodigrim I hope I am not interrupting your holidays ("next week"), requesting a review.

Bodigrim:
@rockbmb sorry for delay, unfortunately, I am still out of time. I hope to regain pace this or next weekend.

test-suite/Math/NumberTheory/ArithmeticFunctionsTests.hs
```diff
@@ -270,6 +272,45 @@ mangoldtOeis = oeisAssertion "A014963" expMangoldtA
   , 73, 1, 1, 1, 1, 1, 79, 1, 3, 1, 83, 1, 1, 1, 1, 1, 89, 1, 1, 1, 1, 1, 1
   ]
 
+nFreedomProperty1 :: Word -> NonZero Natural -> Bool
+nFreedomProperty1 n (NonZero m) =
+    isNFree n m == (all ((< n) . snd) . factorise) m
+
+nFreedomProperty2 :: NonZero Word -> NonNegative Int -> Bool
```
Bodigrim: There is `Math.NumberTheory.TestUtils.Wrappers.Power`, which seems more suitable than `NonZero` here.

test-suite/Math/NumberTheory/ArithmeticFunctionsTests.hs
```diff
@@ -270,6 +272,45 @@ mangoldtOeis = oeisAssertion "A014963" expMangoldtA
...
+nFreedomProperty2 (NonZero n) (NonNegative m) =
+    let n' | n == maxBound = n
+           | otherwise     = n + 1
+    in take m (filter (isNFree n') [1 ..]) == take m (nFrees n' :: [Integer])
+
+nFreedomProperty3 :: NonZero Word -> Positive Int -> Bool
```
Bodigrim: There is `Math.NumberTheory.TestUtils.Wrappers.Power`, which seems more suitable than `NonZero` here.

test-suite/Math/NumberTheory/ArithmeticFunctionsTests.hs
```diff
@@ -270,6 +272,45 @@ mangoldtOeis = oeisAssertion "A014963" expMangoldtA
...
+nFreedomProperty3 (NonZero n) (Positive m) =
+    let n' | n == maxBound = n
+           | otherwise     = n + 1
+        zet = 1 / zetas 1e-14 !! (fromIntegral n') :: Double
+        m' = 100 * m
+        nfree = fromIntegral m' / fromIntegral (last (take m' $ nFrees n' :: [Integer]))
```
Bodigrim: `last . take n` will traverse list twice and thus retain it in memory. Can we use `head . drop (n - 1)` here?

test-suite/Math/NumberTheory/ArithmeticFunctionsTests.hs
```diff
@@ -270,6 +272,45 @@ mangoldtOeis = oeisAssertion "A014963" expMangoldtA
...
+    in 1 / fromIntegral m >= abs (zet - nfree)
+
+-- |
+-- * Using a bounded integer type like @Int@ instead of @Integer@ here means
+-- even a relatively low value of @n@, e.g. 20 may cause out-of-bounds memory
+-- accesses in @nFreesBlock@.
+-- * Using @Integer@ prevents this, so that is the numeric type used here.
+nFreesBlockProperty1 :: NonZero Word -> Positive Integer -> Word -> Bool
```
Bodigrim: There is Math.NumberTheory.TestUtils.Wrappers.Power, which seems more suitable than NonZero here.

Bodigrim:
Sorry for a huge delay. Overall looks good, couple of minor issues in tests.

rockbmb:
@Bodigrim Fixed :+1:

Bodigrim:
Superb! Thanks for your patience.

rockbmb:
@Bodigrim no problem, thanks for reviewing this.

PR #146, opened 2018-10-01
[#144] Fix typo: "Recurrencies" -> "Recurrences"

### LINKED ISSUE #144: Spelling Bug in Module Name Math.NumberTheory.Recurrencies
While "currency" pluralizes to "currencies," "recurrence" pluralizes to "recurrences."  See [Wiktionary](https://en.wiktionary.org/wiki/recurrence).  English is not my native language, so whoever misspelled this has not only my sympathy but shared pain at the intricacies of English spelling.

That said, a fix, if deemed worthwhile, should probably keep the old modules around, just importing and re-exporting the original content with some depreciation warning for some time.

rockbmb:
Closes #144. ~Still WIP.~


rockbmb:
Ok, good to go.

rockbmb:
@Bodigrim this isn't as trivial as the PR from earlier today that only fixes documentation, but please go over it.

Math/NumberTheory/Recurrences/Linear.hs
```diff
@@ -0,0 +1,14 @@
+-- |
+-- Module:      Math.NumberTheory.Recurrences.Pentagonal
```
Bodigrim: Module name is wrong.

Math/NumberTheory/Recurrences/Bilinear.hs
```diff
@@ -0,0 +1,14 @@
+-- |
+-- Module:      Math.NumberTheory.Recurrences.Pentagonal
```
Bodigrim: Module name is wrong.

rockbmb:
@Bodigrim good to go.

Bodigrim:
Thanks!

Issue #79, opened 2017-10-24
Migrate from array to vector


Bodigrim:
Older parts of `arithmoi` tend to use `Data.Array` instead of `Data.Vector`, as newer modules do. It would be nice to migrate to vectors completely. The migration should be pretty straightforward, so it is a nice first issue.

tau3:
Hello! 

I'd like to try and implement this feature if nobody is currently working on it. Could you give me a hint on `castSTUArray` - what is the similar function for `STVector`?

Bodigrim:
`castSTUArray` is basically `return . unsafeCoerce`, no more no less. But in fact we should get rid of `castSTUArray :: STUArray s Int Bool -> ST s (STUArray s Int Word))` at all. 

The thing is that this code in `arithmoi` is based upon a assumption that `Array a Bool` is a tightly-packed bit sequence, 8 `Bool` per byte, 32/64 `Bool` per machine word. This assumption is not guaranteed in general and, what is more important, is wrong for vectors: `Vector Bool` stores one `Bool` per byte, 4/8 `Bool` per machine word. 

That said, one should replace 
```haskell
data PrimeSieve = PS !Integer {-# UNPACK #-} !(UArray Int Bool)
```
by
```haskell
data PrimeSieve = PS !Integer {-# UNPACK #-} !(UArray Int Word)
```
and take care of correct indexing of individual bits. This is better to be done in a separate small PR, before taking a stab to current issue. 

---

I believe @rockbmb had plans to fix this issue. @rockbmb are you interested to do it yourself or @tau3 may take over?

tau3:
@Bodigrim 
Thanks for explanation!

@rockbmb 
I have already refactored `MoebiusInversion` and `Cubes`, please take it if you need 😄

rockbmb:
@tau3 I have some code that may be useful to the issue, I'll push it to a branch in my fork and you can see if it'll help you.

Bodigrim:
I picked some low-hanging fruits in 9c901536c713fd96e94fc7c2237c39ada2cb88c7 and b6e114511bfcf5cabf83f4e157d0940d422bbbd5. See #157 for discussion about vectors for sieves.

rockbmb:
@Bodigrim did this change introduce any changes in benchmarks? I've always been curious as to which is generally faster, `Array` or `Vector`, although this example might be too simple to give correct answers.

Bodigrim:
We do not have benchmarks for these parts of the project. 

[...]

PR #147, opened 2018-10-18
WIP: Migration to vector

tau3:
Hello!

Here is the initial implementation of migration from arrays to vectors  (#79) It compiles but fails against some of unit tests. To figure out the root cause, i recompiled sources with enabled `check-bounds` Yep, seems like most of exceptions were 'index of bounds'. I wanted to debug in `cabal-repl`, but it failed to load due to `Error: bytecode compiler can't handle unboxed tuples and sums.` Cabal suggested to specify `-fobject-code`, but i couldn't google anything about that flag. 

So, some assistance is required :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ stack.yaml
 dist
 dist-newstyle
 .ghc.environment*
+.cabal-sandbox
+cabal.sandbox.config
+
--- a/Math/NumberTheory/MoebiusInversion.hs
+++ b/Math/NumberTheory/MoebiusInversion.hs
@@ -17,9 +17,7 @@ import Control.Monad
 import Control.Monad.ST
 
 import Math.NumberTheory.Powers.Squares
-
-import Data.Vector as V (freeze, unsafeIndex)
-import Data.Vector.Mutable as MV (STVector, new, unsafeRead, unsafeWrite)
+import Math.NumberTheory.Unsafe
 
 -- | @totientSum n@ is, for @n > 0@, the sum of @[totient k | k <- [1 .. n]]@,
 --   computed via generalised Möbius inversion.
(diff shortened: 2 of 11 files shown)


tau3:
P.S. diff shows realy huuuuge changeset because of hindent (default setup). If a special profile is used in Arithmoi, please, point me at so i can reduce the number of changed lines.

rockbmb:
@tau3 when you run `cabal repl` or `cabal new-repl`, simply pass `-fobject-code` as a GHC option:

```
cabal (new-)repl --ghc-options=-fobject-code
```

This goes for any other GHC option(s) you may wish to pass to a GHCi session.

Bodigrim:
> Cabal suggested to specify -fobject-code, but i couldn't google anything about that flag.

https://downloads.haskell.org/~ghc/8.4.1/docs/html/users_guide/ghci.html#compiling-to-object-code-inside-ghci

> P.S. diff shows realy huuuuge changeset because of hindent (default setup). If a special profile is used in Arithmoi, please, point me at so i can reduce the number of changed lines.

I do not use `hindent` for `artithmoi`, so no agreed profile exists. We can discuss indentation profile as a separate issue, but in course of this PR I am not keen to review thousand lines. Please commit only non-whitespace changes. You can do it as suggested at https://stackoverflow.com/a/7149602 plus several manual tweaks.

tau3:
@Bodigrim thanks for advices, but unfortunatelly `git diff ... -U0 -w ...` didn't do the trick with spaces in this case :disappointed: 

also i encountered some weird vector bounds errors during debug (debugging itself was kinda stressing due to inability to work with unboxed code in interpreter mode).

probably this refactoring attempt should be considered as unseccessful.

@rockbmb thank you and good luck with yours!)

rockbmb:
@tau3 don't call it unsuccessful just yet! You don't have to do everything at once.

You can just pick a module that is all four of:
* small,
* simple
* few other modules depend on it
* is still using `Array`s instead of `Vector`s

and make the necessary refactoring. It won't be too difficult, and remember that after it compiles you have a friend in `cabal (new-)test`.

Bodigrim:
+1 to @rockbmb 
It may be tough to migrate prime sieves from `array` to `vector`, but migration of other modules is also highly appreciated.

Bodigrim:
Closing due to the lack of activity. Anyway thanks for you efforts!

PR #150, opened 2018-12-24
[#149] Create initial version of 'partition' function

### LINKED ISSUE #149: Computing the partition function for a single value
After #107, this library now offers a way to calculate a contiguous sequence of values of the [partition function](https://en.wikipedia.org/wiki/Partition_function_(number_theory)).

However, for a large, isolated `n`, this is impractical because all values up to this `n` have to be calculated, and as stated [here](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/Recurrences/Pentagonal.hs#L68), it is not possible to calculate beyond the `2^64`-th element of this (zero-indexed) sequence (on 64 bit machines).

For the purpose of calculating `p(n)` for a single value, the [Hardy-Ramanujan-Rademacher formula](https://arxiv.org/abs/1205.5991) exists.

This issue will be about writing a new function `partitionOf :: Integral a => Natural -> a` (the type is most likely wrong and the name can be made more suggestive later).

@Bodigrim would you find this worthwhile?

rockbmb:
Closes #149. No tests, no benchmarks, no comments, and incorrect results. However, it works, and I'll be improving the code with these things. Ideally I reproduce the author's `p(10^19)` result.

- [x] Make it work with `Double` (or more generally, `Floating`).
- [ ] Improve the function so that `p(n)` does return a result with `O(n^2)` bits of precision. Some type-level programming will be required to return a fixed-point type whose precision is dependent on the input.
- [ ] Write tests that verify (should I put "verify" between quotes :smiley:) the function's correctness
- [ ] Write benchmarks
- [ ] Write better comments

### DIFF
--- a/Math/NumberTheory/Partition.hs
+++ b/Math/NumberTheory/Partition.hs
@@ -14,13 +14,51 @@ module Math.NumberTheory.Partition
   ( partition
   ) where
 
-import Math.NumberTheory.Powers.Squares (integerSquareRoot)
+import qualified Math.NumberTheory.Recurrences as R
+import Numeric.Rounded (Rounding (..), Precision (..))
 
 partition :: forall a b . (Integral a, Floating b) => a -> b
-partition num = sum . map (_T num) $ [1 .. _N]
+partition num | num < 128 = R.partition !! (fromIntegral num)
+              | otherwise = sum . map (_T num) $ [1 .. _N]
   where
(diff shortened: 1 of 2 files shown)


rockbmb:
*Very* stale and honestly, not really sure this is a good idea anymore.

Should this ever become a necessary feature, one can just come back to this PR and reuse whatever is possible.

PR #158, opened 2019-01-08
(WIP) Improved Cyclic group type

### LINKED ISSUE #154: Improve return type of isPrimitiveRoot'
Since `PrimitiveRoot a` is an abstract data type with a hidden constructor (#130), function `isPrimitiveRoot' :: CyclicGroup a -> a -> Bool` is pretty much useless: even if it returns `True`, there are no means to use this boolean as an evidence to construct a `PrimitiveRoot` value. And the only way to get `PrimitiveRoot` is using `isPrimitiveRoot :: Mod n -> Maybe (PrimitiveRoot n)`, which runs (possibly expensive) `cyclicGroupFromModulo` for each candidate. 

I do not have any good ideas at the moment, how to fix this.

b-mehta:
A very rough draft for a possible fix for #154.
`isPrimitiveRoot'2` is essentially just a smart constructor for `PrimitiveRoot` now, while `isPrimitiveRoot2` has virtually the same functionality as before.
In addition, `isPrimitiveRoot'2` avoids the potentially expensive factorisation of modulus.

Possibly fixes #154.

### DIFF
--- a/Math/NumberTheory/Moduli/PrimitiveRoot.hs
+++ b/Math/NumberTheory/Moduli/PrimitiveRoot.hs
@@ -8,12 +8,14 @@
 --
 
 {-# LANGUAGE CPP                  #-}
+{-# LANGUAGE DataKinds            #-}
 {-# LANGUAGE DeriveGeneric        #-}
 {-# LANGUAGE FlexibleContexts     #-}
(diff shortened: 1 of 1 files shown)


Bodigrim:
Closed in favor of #169.

PR #168, opened 2019-06-09
Separate Euclidean to its own package

### LINKED ISSUE #166: Separate Euclidean to its own package
Type class `Integral`  is completely messed up. 

```haskell
class (Real a, Enum a) => Integral a where 
  toInteger :: a -> Integer 
  quotRem   :: a -> a -> (a, a)
  ...

class (Num a, Ord a) => Real a where 
  toRational :: a -> Rational 
```

First of all, `Real` constraint looks really weird. How is it connected to being `Integral`? 

Having both `toInteger :: a -> Integer` and `Num.fromInteger :: Integer -> a` means that `a` is isomorphic to a subset of `Integer`.

(Moreover, `Enum.toEnum :: a -> Int` and `Enum.fromEnum :: a -> Int` imply that `a` is no greater than `Int`, which is untrue even for `Integer` and `Natural`, but who cares).

This is very restrictive, because there are many rings, larger than `Integer`, where one can still define a sensible `quotRem`. And having `quotRem` is enough for `gcd`, `lcm`, etc. Obvious example is a ring of polynomials, together with its factor rings (Gaussian integers, Eisenstein integers, etc.).

For the reasons above @rockbmb and I developed a type class `Euclidean` (https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/Euclidean.hs), representing [Euclidean domains](https://en.wikipedia.org/wiki/Euclidean_domain). It is basically `Integral` without `toInteger`:

```
class (Eq a, Num a) => Euclidean a where 
  quotRem   :: a -> a -> (a, a)
  ...  
```

It is used extensively throughout `arithmoi` for [quadratic integers](https://en.wikipedia.org/wiki/Quadratic_integer). 

--- 

I want to use `Euclidean` for polynomials in https://github.com/Bodigrim/poly, but `arithmoi` itself is a quite heavy dependency. So my intention is to extract it into a dedicated package.

Bodigrim:
Closes #166 and (accidentally) #167.

### DIFF
--- a/Math/NumberTheory/ArithmeticFunctions/Inverse.hs
+++ b/Math/NumberTheory/ArithmeticFunctions/Inverse.hs
@@ -9,6 +9,7 @@
 -- <https://www.emis.de/journals/JIS/VOL19/Alekseyev/alek5.pdf Computing the Inverses, their Power Sums, and Extrema for Euler’s Totient and Other Multiplicative Functions>
 -- by M. A. Alekseyev.
 
+{-# LANGUAGE CPP                 #-}
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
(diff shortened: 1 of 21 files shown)


rockbmb:
Very nice.

PR #172, opened 2019-09-24
WIP Atkin sieve

### LINKED ISSUE #157: Implement the sieve of Atkin
[Our implementation](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/Primes/Sieve/Eratosthenes.hs) of Eratosthenes sieve is notoriously difficult to read. It has been heavily optimized long time ago and I have no clue if these patterns still make sense for modern GHCs and modern PCs. It is also quite possible that bugs are lurking in dark corners. I tried to disentangle it a couple of times, but always retreated in horror. Basically, it is too easy to introduce an off-by-one error, which will manifest itself only for a small portion of cases, possibly very large ones. 

Maybe there is a better path: reimplement prime sieves altogether using [the sieve of Atkin](https://en.wikipedia.org/wiki/Sieve_of_Atkin). It has better performance characteristics and the relevant [paper](http://www.ams.org/journals/mcom/2004-73-246/S0025-5718-03-01501-1/S0025-5718-03-01501-1.pdf) is super clear. 

---

*Note on containers.* Current implementation represents sieves by `UArray Int Bool`. This is clearly the wrong level of abstraction: since we do not employ fancy indexing via `Ix` typeclass, we'd rather stick either to low-level arrays from `primitive` or to full-fledged vectors from `vector`. The arrays in question are quite long and we "grow" them in [`growCache`](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/Primes/Sieve/Eratosthenes.hs#L293), so `Vector` is the right choice. 

Unfortunately, vanilla `Vector` does not support dense bit arrays (64 bits per `Word64`). See also https://github.com/haskell/primitive/issues/42. There is [`bitvec`](http://hackage.haskell.org/package/bitvec) package, which aims to provide `Vector` interface to dense bitmaps, but it is abandoned: [known issues](https://github.com/mokus0/bitvec/issues/4) are open for two years and the test suite does not build anymore. So for the time being there is no other choice but continue using `UArray Int Bool`.

Bodigrim:
Closes #157

It is already competitive with the existing sieve and is 10 times simpler, but I need to merge more performance patches into https://github.com/Bodigrim/bitvec to make this code really shine.

### DIFF
--- a/Math/NumberTheory/Primes.hs
+++ b/Math/NumberTheory/Primes.hs
@@ -20,6 +20,9 @@ module Math.NumberTheory.Primes
     , factorBack
     , -- * Old interface
       primes
+    , -- * Temporary
+      atkinPrimeList
+    , atkinSieve
     ) where
 
 import Data.Bits
@@ -31,6 +34,7 @@ import Numeric.Natural
 import Math.NumberTheory.Primes.Counting (nthPrime, primeCount)
 import qualified Math.NumberTheory.Primes.Factorisation.Montgomery as F (factorise)
 import qualified Math.NumberTheory.Primes.Testing.Probabilistic as T (isPrime)
+import Math.NumberTheory.Primes.Sieve.Atkin
 import Math.NumberTheory.Primes.Sieve.Eratosthenes (primes, sieveRange, primeList, psieveFrom, primeSieve)
 import Math.NumberTheory.Primes.Small
 import Math.NumberTheory.Primes.Types
--- a/Math/NumberTheory/Primes.hs
+++ b/Math/NumberTheory/Primes.hs
@@ -21,8 +21,7 @@ module Math.NumberTheory.Primes
     , -- * Old interface
       primes
     , -- * Temporary
-      atkinPrimeList
-    , atkinSieve
+      module Math.NumberTheory.Primes.Sieve.Atkin
     ) where
 
 import Data.Bits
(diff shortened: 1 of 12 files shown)


Bodigrim:
This is more or less ready _per se_. But unfortunately prime counting is so intimately (and impenetrably) connected to the existing Eratosthenes sieve, that I cannot decomission the latter. That said, this PR is on hold until I disentangle this knot.

rockbmb:
Hey @Bodigrim,

> unfortunately prime counting is so intimately (and impenetrably) connected to the existing Eratosthenes sieve, that I cannot decomission the latter.

I understand that merging this as is without removing the earlier sieve code would just clutter the codebase (even further), but if it's already working with good performance, wouldn't merging it now and making a new PR that proceeds with decoupling the `Math.NumberTheory.Primes.Counting` API from `Math.NumberTheory.Primes.Sieve.Eratosthenes` also work?

Bodigrim:
My main goal here is to make source code more maintainable. I can even agree to worsen performance mildly. 

It might happen that I'd never untangle `Math.NumberTheory.Primes.Counting`. Having two implementations of sieves side-by-side is not something future maintainers would be thankful for ;)

rockbmb:
@Bodigrim can't say I don't understand. It's already difficult enough maintaining the package as it is - I remember trying to migrate from `array` to `vector`, that was challenging.

Issue #155, opened 2018-12-30
Cubic reciprocity law


Bodigrim:
It may be fun to extend [Math.NumberTheory.Moduli.Jacobi](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/Moduli/Jacobi.hs) to cover not only [quadratic reciprocity](https://en.wikipedia.org/wiki/Quadratic_reciprocity), but [cubic](https://en.wikipedia.org/wiki/Cubic_reciprocity) as well.

Quadratic reciprocity symbol or [Jacobi symbol](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/Moduli/Jacobi.hs#L33) answers the question whether an equation `x^2 = n (mod m)` has any solutions. Similarly, cubic residue symbol answers whether `x^3 = n (mod m)` has solutions.

The values of Jacobi symbol are units of real integers (+ zero), while the values of cubic symbol are units of Eisenstein integers.

When this will be implemented, we can proceed to modular cubic roots (similar to [Math.NumberTheory.Moduli.Sqrt](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/Moduli/Sqrt.hs)) and eventually solve cubic and quartic modular equations in [Math.NumberTheory.Moduli.Equations](https://github.com/cartazio/arithmoi/blob/master/Math/NumberTheory/Moduli/Equations.hs).

hgsipiere:
I was looking into this issue out of curiosity and noticed that with cubic/quartic residue characters the values are roots of unity. Would it make sense to move roots of unity out of Dirichlet characters or to just reference the Dirichlet character module if implementing this?

Edit: Ignore this, can just use EisensteinIntegers module

Bodigrim:
Moving them out is a good idea, I think.

b-mehta:
I agree roots of unity probably belong somewhere other than in Dirichlet characters, but cubic/quartic symbols should have a more specific type than root of unity. It would be nice to include a conversion function `CubicResidueSymbol -> RootOfUnity` or something similar instead.

hgsipiere:
I completely agree, representing them as Eisenstein Integers looks nice with stuff like CubicResidueSymbol ω (or anything else more specific). It was an oversight, I think they are seperate issues.

Edit: I think you can only do CubicResidueSymbol -> Maybe RootOfUnity because of the 0 case.

Bodigrim:
[...]

PR #180, opened 2020-01-05
Dirichlet characters

b-mehta:
- [x] Create `M.NT.Moduli.Internal` module, for `discreteLogarithmPP` and `isPrimitiveRoot'`, so we can work with these internally without needing to use `CyclicGroup`.
- [x] I created `Template` a while ago, but since then `SFactors` was introduced - this can be adapted to make a nicer version of `Template`. It's unclear if `Template` should have the modulus on type level though, considering `induced` might be easier to write without that. Edit: This turned out to be a bad idea, due to the reason mentioned.
- [x] Figure out how to do `isPrimitive`
- [x] Bulk evaluation test currently fails. The [branch here](https://github.com/b-mehta/arithmoi/tree/dirichlet-characters-wip) might have the fix for this.


Bodigrim:
This looks awesome! Extra kudos for beautiful TeX formulas.

Bodigrim:
(Just random notes)
At some point we would probably need a separate module dedicated to the roots of unity. There are `Moebius` and `JacobiSymbol`, which effectively are just `OrZero (RootsOfUnity 2)`, cubic residue symbols (#155) are `OrZero (RootsOfUnity 3)`, and units of Gaussian and Eisenstein integers are `RootsOfUnity 4` and `RootsOfUnity 6`.

Bodigrim:
BTW `OrZero a` seems to be equivalent to `Data.Monoid.Ap Maybe a`. I would rather stick with the latter, possibly adding a synonym `type OrZero = Ap Maybe`.

b-mehta:
> At some point we would probably need a separate module dedicated to the roots of unity. 

Yeah this is a good point, but I'm hesitant to make the RootOfUnity here be parametrised on type level. In particular, it would make
```haskell
evaluate :: DirichletCharacter n -> MultMod n -> RootOfUnity (Phi n)
```
which would be a pain (I did try). Perhaps `Some RootsOfUnity` would be better, but it seems a little awkward to work with.

Bodigrim:
Sure, let us leave `RootOfUnity` as is, at least for now.

b-mehta:
> BTW OrZero a seems to be equivalent to Data.Monoid.Ap Maybe a. I would rather stick with the latter, possibly adding a synonym type OrZero = Ap Maybe.

Ah, thanks for this! I'll convert to this, possibly using some pattern synonyms as well

b-mehta:
Actually, `Data.Monoid.Ap` was introduced in base 4.12, which would be a fairly strong new constraint... Should I still use it, or is it best to use `OrZero` for now, and create an issue to convert to `Data.Monoid.Ap` in the future?

Bodigrim:
I'd use `Data.Monoid.Ap` and define it for earlier versions of `base` along these lines:

```haskell
#if !MIN_VERSION_base(4,12,0)
newtype Ap f a = Ap { getApp :: f a }
  deriving (Functor, Applicative, Monad)

instance (Applicative f, Semigroup a) => Semigroup (Ap f a) where
  (<>) = liftA2 (<>)

instance (Applicative f, Monoid a) => Monoid (Ap f a) where
  mempty = pure mempty
  mappend = (<>)
#endif 
```

b-mehta:
I need to write a quick test for `orderChar`, but other than that I think this is essentially ready to review now. There's a few TODOs in there, but they can be separate issues so that this is easier to review.

b-mehta:
Do you have a preference for a Show instance for `DirichletCharacter`? The options as I see them are:

- None
- "χ_r with modulus n"
- Enumerate its values
- The internal representation
- Another option?

It's also tempting to implement `OrZero` with newtypes, so that a custom Show instance can be given, do you think this is a good idea?

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,589 @@
+-- |
+-- Module:      Math.NumberTheory.DirichletCharacters
+-- Copyright:   (c) 2018 Bhavik Mehta
+-- Licence:     MIT
+-- Maintainer:  Bhavik Mehta <bhavikmehta8@gmail.com>
+-- Stability:   Provisional
+-- Portability: Non-portable (GHC extensions)
```
Bodigrim: I usually ignore `Stability` and `Portability`, IMHO they do not make much sense nowadays.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,589 @@
...
+
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE CPP #-}
```
Bodigrim: Please sort pragmas alphabetically and justify right ends.

arithmoi.cabal
```diff
@@ -59,13 +59,15 @@ library
...
     Math.NumberTheory.Curves.Montgomery
+    Math.NumberTheory.DirichletCharacters
     Math.NumberTheory.Euclidean
     Math.NumberTheory.Euclidean.Coprimes
     Math.NumberTheory.Moduli
     Math.NumberTheory.Moduli.Chinese
     Math.NumberTheory.Moduli.Class
     Math.NumberTheory.Moduli.DiscreteLogarithm
     Math.NumberTheory.Moduli.Equations
+    Math.NumberTheory.Moduli.Internal
```
Bodigrim: I believe internals are not intended to be publicly exposed. Please move them under `other-modules` section.

Math/NumberTheory/Moduli/JacobiSymbol.hs
```diff
@@ -48,6 +49,18 @@ negJS = \case
   Zero     -> Zero
   One      -> MinusOne
 
+{-# SPECIALISE symbolToIntegral :: JacobiSymbol -> Integer,
+                                   JacobiSymbol -> Int,
+                                   JacobiSymbol -> Word,
+                                   JacobiSymbol -> Natural
+  #-}
+-- | Convenience function to convert out of a Jacobi symbol
+symbolToIntegral :: Integral a => JacobiSymbol -> a
```
Bodigrim: Probably `Num` constraint would be enough.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,589 @@
...
+-- | Induce a Dirichlet character to a higher modulus. If \(d \mid n\), then \(a \bmod{n}\) can be
+-- reduced to \(a \bmod{d}\). Thus, the multiplicative function on \(\mathbb{Z}/d\mathbb{Z}\)
+-- induces a multiplicative function on \(\mathbb{Z}/n\mathbb{Z}\).
+--
+-- >>> :set -XTypeApplications
+-- >>> chi = indexToChar 5 :: DirichletCharacter 45
+-- >>> chi2 = induced @135 chi
+-- >>> :t chi2
+-- Maybe (DirichletCharacter 135)
+induced :: forall n d. (KnownNat d, KnownNat n) => DirichletCharacter d -> Maybe (DirichletCharacter n)
```
Bodigrim, line 383: If you add `Mod n d ~ 0` to constraints, it would be possible to avoid `Maybe` at all.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,589 @@
...
+  where primitive :: DirichletFactor -> Bool
+        primitive Two = False
+        -- for odd p, we're testing if phi(p^(k-1)) `stimes` a is 1, since this means the
+        -- character can come from some the smaller modulus p^(k-1)
+        primitive (OddPrime _ 1 _ a) = a /= mempty
+        primitive (OddPrime (unPrime -> p) k _ a) = (p^(k-2)*(p-1)) `stimes` a /= mempty
+        primitive (TwoPower 2 a _) = a /= mempty
+        primitive (TwoPower k _ b) = (bit (k-3) :: Integer) `stimes` b /= mempty
+
+-- | A Dirichlet character is primitive if cannot be @induced@ from any character with
```
Bodigrim: Should be `'induced'`, not `@induced@`.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,589 @@
...
+  -- ** Construction
+  , indexToChar
+  , indicesToChars
+  , characterNumber
+  , allChars
+  -- ** Evaluation
+  , evaluate
+  , generalEval
+  , toFunction
+  , allEval
```
Bodigrim: Let us use the same stem: either `eval`, `generalEval`, `allEval` or `evaluate`, `generalEvaluate`, `allEvaluate`. I would also put the verb first: e. g., `evalAll` and `evalGeneral`.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,589 @@
...
+        correctPrimitiveRoot (OddPrime p k g _) = g == generator p k
+        correctPrimitiveRoot _ = True
+        validValued (TwoPower k a b) = a <> a == mempty && (bit (k-2) :: Integer) `stimes` b == mempty
+        validValued (OddPrime (unPrime -> p) k _ a) = (p^(k-1)*(p-1)) `stimes` a == mempty
+        validValued Two = True
+        n = natVal (Proxy :: Proxy n)
+        two = toEnum 1 -- lazy way to get Prime 2
+
+-- | Get the order of the Dirichlet Character.
+-- TODO: test this
```
Bodigrim: Remove `TODO`.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,589 @@
...
+        getPP (TwoPower k _ _) = (two, fromIntegral k)
+        getPP (OddPrime p k _ _) = (p, k)
+        getPP Two = (two,1)
+        correctPrimitiveRoot (OddPrime p k g _) = g == generator p k
+        correctPrimitiveRoot _ = True
+        validValued (TwoPower k a b) = a <> a == mempty && (bit (k-2) :: Integer) `stimes` b == mempty
+        validValued (OddPrime (unPrime -> p) k _ a) = (p^(k-1)*(p-1)) `stimes` a == mempty
+        validValued Two = True
+        n = natVal (Proxy :: Proxy n)
+        two = toEnum 1 -- lazy way to get Prime 2
```
Bodigrim: Better use `minBound` :) 
It is actually an oversight from my side that `toEnum 1` is 2 and not 3.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,589 @@
...
+  , characterNumber
+  , allChars
+  -- ** Evaluation
+  , evaluate
+  , generalEval
+  , toFunction
+  , allEval
+  -- ** Special Dirichlet characters
+  , principalChar
+  , isPrincipal
```
Bodigrim, line 45: Isn't it redundant to export `isPrincipal`, when we have `Eq` instance already available? I would rather keep the export list as laconic as practical.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,589 @@
...
+  (
+  -- * Roots of unity
+    RootOfUnity
+  -- ** Conversions
+  , toRootOfUnity
+  , fromRootOfUnity
+  , toComplex
+  -- * An absorbing semigroup
+  , OrZero, pattern Zero, pattern NonZero
+  , asNumber
```
Bodigrim: I would probably do not export this function. It is just `maybe 0`, not worth of polluting global namespace with a rather ambiguous name.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,589 @@
...
+  show (RootOfUnity q)
+    | n == 0    = "1"
+    | d == 1    = "-1"
+    | n == 1    = "e^(πi/" ++ show d ++ ")"
+    | otherwise = "e^(" ++ show n ++ "πi/" ++ show d ++ ")"
+    where n = numerator (2*q)
+          d = denominator (2*q)
+
+-- | Given a rational \(q\), produce the root of unity \(e^{2 \pi i q}\).
+toRootOfUnity :: Integral a => Ratio a -> RootOfUnity
```
Bodigrim: I'd rather avoid excessively polymorphic functions. Each call of `toRootOfUnity (2 % 3)` raises `Defaulting the following constraints to type ‘Integer’` warning. 

If you cast an argument to `Rational` first thing and do not use type `a` in any way, then it is better to make `toRootOfUnity` monomorphic, accepting only `Rational`. Users can easily combine it with `toRational` themselves, when needed.

Bodigrim:
> Do you have a preference for a Show instance for `DirichletCharacter`? 

I would probably just dump its internal representation, but no strong opinion here.

> It's also tempting to implement `OrZero` with newtypes, so that a custom Show instance can be given, do you think this is a good idea?

Not realy. Too many layers and new newtypes make users struggle. However, you can probably define `instance {-# OVERLAPPING #-} Show (Ap (Maybe RootOfUnity))`.

Math/NumberTheory/DirichletCharacters.hs
b-mehta: Fair enough, I will remove.

arithmoi.cabal
b-mehta: Thanks for this catch!

Math/NumberTheory/DirichletCharacters.hs
b-mehta: `minBound` doesn't work here because we would like a `Prime Natural`, which is not `Bounded`.

Math/NumberTheory/DirichletCharacters.hs
b-mehta: Not really. I think your idea is that `isPrincipal chi` should be equivalent to `chi == mempty`, but `mempty` requires a `KnownNat n` instance, while `isPrincipal` does not. 
(`mempty` should require this instance, since to create a character from thin air, we need to know the modulus on term level)

Math/NumberTheory/DirichletCharacters.hs
b-mehta: It's `maybe 0` but with the newtype abstraction `Ap` in the way, so this would be useful to export. I'll happily take better name suggestions.

Math/NumberTheory/DirichletCharacters.hs
Bodigrim: Makes sense, thanks.

Math/NumberTheory/DirichletCharacters.hs
Bodigrim: Ahh, I see. What about `nextPrime 2`?

b-mehta:
I would consider using overlapping instances significantly worse, since this extension can lead to incoherence. I think I'll just avoid a Show instance altogether.

Math/NumberTheory/DirichletCharacters.hs
b-mehta: Amended, thanks.

Math/NumberTheory/DirichletCharacters.hs
b-mehta: Adjusted, thanks.

Math/NumberTheory/DirichletCharacters.hs
b-mehta: Thanks.

Math/NumberTheory/DirichletCharacters.hs
b-mehta: I switched to this, since I had no strong preference either way.

Math/NumberTheory/DirichletCharacters.hs
b-mehta: Changed, thanks

Bodigrim:
> Yeah this is a good point, but I'm hesitant to make the RootOfUnity here be parametrised on type level. In particular, it would make
>
> ```haskell
> evaluate :: DirichletCharacter n -> MultMod n -> RootOfUnity (Phi n)
> ```
>
> which would be a pain (I did try).

Actually I think it is possible to avoid `Phi` along these lines:

```haskell
-- | RootOfUnity m :: RootOfUnity n represents \(e^{2 \pi i m / \phi(n)}\).
newtype RootOfUnity n = RootOfUnity Integer
  deriving (Eq)

instance KnownNat n => Show (RootOfUnity n) where
  show (RootOfUnity m) = "e^(" ++ show (numerator r) ++ "πi/" ++ show (denominator r) ++ ")"
    where
      r = 2 * m % totient (natVal (Proxy :: Proxy n))

toComplex :: Floating a => SFactors Integer n -> RootOfUnity n -> Complex a
toComplex fact (RootOfUnity m) = cis (2 * pi * m / d)
  where
    d = runFunctionOnFactors totientA (unSFactors fact)

evaluate :: DirichletCharacter n -> MultMod n -> RootOfUnity n
evaluate = undefined
```

(I do not mean this to be done in this branch)

Math/NumberTheory/Moduli/Internal.hs
```diff
@@ -0,0 +1,162 @@
+-- |
+-- Module:      Math.NumberTheory.Moduli.Multiplicative
+-- Copyright:   (c) 2017 Andrew Lelechenko
```
Bodigrim: Please update `Module` and `Copyright` fields.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,588 @@
...
+    | n == 1    = "e^(πi/" ++ show d ++ ")"
+    | otherwise = "e^(" ++ show n ++ "πi/" ++ show d ++ ")"
+    where n = numerator (2*q)
+          d = denominator (2*q)
+
+-- | Given a rational \(q\), produce the root of unity \(e^{2 \pi i q}\).
+toRootOfUnity :: Rational -> RootOfUnity
+toRootOfUnity q = RootOfUnity ((n `rem` d) % d)
+  where n = toInteger $ numerator q
+        d = toInteger $ denominator q
```
Bodigrim: Both casts are redundant.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,588 @@
...
+-- | A character can evaluate to a root of unity or zero: represented by @Nothing@.
+evalGeneral :: KnownNat n => DirichletCharacter n -> Mod n -> OrZero RootOfUnity
+evalGeneral chi t = case isMultElement t of
+                      Nothing -> Zero
+                      Just x -> NonZero $ eval chi x
+
+-- | Convert a Dirichlet character to a complex-valued function. As in `toComplex`, the result is
+-- inexact due to floating-point inaccuracies. See `toComplex`.
+toFunction :: (Integral a, RealFloat b, KnownNat n) => DirichletCharacter n -> a -> Complex b
+toFunction chi = asNumber toComplex . evalGeneral chi . fromIntegral
```
Bodigrim: Similar to discussion about `toRootOfUnity`, please make this function monomorphic in `a`, accepting `Mod n` only.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,588 @@
...
+
+-- | Give the dirichlet character from its number.
+-- Inverse of `characterNumber`.
+indexToChar :: forall n a. (KnownNat n, Integral a) => a -> DirichletCharacter n
+indexToChar = runIdentity . indicesToChars . Identity
+
+-- | Give a collection of dirichlet characters from their numbers. This may be more efficient than
+-- `indexToChar` for multiple characters, as it prevents some internal recalculations.
+indicesToChars :: forall n a f. (KnownNat n, Integral a, Functor f) => f a -> f (DirichletCharacter n)
+indicesToChars = fmap (Generated . unroll t . (`mod` m) . fromIntegral)
```
Bodigrim: Again, just consume `f Natural` instead of `f a` and eliminate `fromIntegral`.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,588 @@
...
+  where m = totient $ natVal (Proxy :: Proxy n)
+
+-- | The same as `indexToChar`, but if we're given a character we can create others more efficiently.
+makeChar :: Integral a => DirichletCharacter n -> a -> DirichletCharacter n
+makeChar x = runIdentity . bulkMakeChars x . Identity
+
+-- | Use one character to make many more: better than indicesToChars since it avoids recalculating
+-- some primitive roots
+bulkMakeChars :: (Integral a, Functor f) => DirichletCharacter n -> f a -> f (DirichletCharacter n)
+bulkMakeChars x = fmap (Generated . unroll t . (`mod` m) . fromIntegral)
```
Bodigrim, line 318: Again, just consume `f Natural` instead of `f a` and eliminate `fromIntegral`.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,588 @@
...
+        real (OddPrime _ _ _ a) = a <> a == mempty
+        real (TwoPower _ _ b) = b <> b == mempty
+        real Two = True
+
+-- TODO: it should be possible to calculate this without eval/evalGeneral
+-- and thus avoid using discrete log calculations: consider the order of m
+-- inside each of the factor groups?
+-- | Evaluate a real Dirichlet character, which can only take values \(-1,0,1\).
+toRealFunction :: (Integral a, KnownNat n) => RealCharacter n -> a -> Int
+toRealFunction (RealChar chi) m = case evalGeneral chi (fromIntegral m) of
```
Bodigrim: Please limit `a` to `Mod n` and eliminate `fromIntegral`.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,588 @@
...
+import Data.Complex                                        (Complex(..), cis)
+import Data.Foldable                                       (for_)
+import Data.Functor.Identity                               (Identity(..))
+import Data.List                                           (mapAccumL, foldl', sort, find)
+import Data.Maybe                                          (mapMaybe)
+#if MIN_VERSION_base(4,12,0)
+import Data.Monoid                                         (Ap(..))
+#endif
+import Data.Proxy                                          (Proxy(..))
+import Data.Ratio                                          (Rational, (%), numerator, denominator)
```
Bodigrim: I'm getting a warning that `The import of ‘Rational’ from module ‘Data.Ratio’ is redundant`.

Math/NumberTheory/DirichletCharacters.hs
```diff
@@ -0,0 +1,588 @@
...
+  , isPrimitive
+  , getPrimitiveChar
+  , induced
+  , makePrimitive
+  , WithNat(..)
+  -- * Debugging
+  , validChar
+  ) where
+
+import Control.Applicative                                 (liftA2)
```
Bodigrim, line 65: Please guard it using `CPP`:

```haskell
#if !MIN_VERSION_base(4,12,0)
import Control.Applicative 
#endif 
```

Bodigrim:
Overall a very nice code. I really appreciate the amount of work and engineering.

b-mehta:
> > Yeah this is a good point, but I'm hesitant to make the RootOfUnity here be parametrised on type level. In particular, it would make
> > ```haskell
> > evaluate :: DirichletCharacter n -> MultMod n -> RootOfUnity (Phi n)
> > ```
> > 
> > 
> > which would be a pain (I did try).
> 
> Actually I think it is possible to avoid `Phi` along these lines:
> 
> ```haskell
> -- | RootOfUnity m :: RootOfUnity n represents \(e^{2 \pi i m / \phi(n)}\).
> newtype RootOfUnity n = RootOfUnity Integer
>   deriving (Eq)
> 
> instance KnownNat n => Show (RootOfUnity n) where
>   show (RootOfUnity m) = "e^(" ++ show (numerator r) ++ "πi/" ++ show (denominator r) ++ ")"
>     where
>       r = 2 * m % totient (natVal (Proxy :: Proxy n))
> 
> toComplex :: Floating a => SFactors Integer n -> RootOfUnity n -> Complex a
> toComplex fact (RootOfUnity m) = cis (2 * pi * m / d)
>   where
>     d = runFunctionOnFactors totientA (unSFactors fact)
> 
> evaluate :: DirichletCharacter n -> MultMod n -> RootOfUnity n
> evaluate = undefined
> ```
> 
> (I do not mean this to be done in this branch)

This would work but it would mean RootOfUnity isn't the ideal name and we cannot represent cubic residues... I attempted a few other versions of this module which would be more typed, but they turned out to be awkward to use - you can see some attempts in my wip branch. I opted instead for a version without putting much on type level, but with tests that it all works (using validChar) instead

Math/NumberTheory/DirichletCharacters.hs
b-mehta: Do you think it is worth keeping this function? I'd put it in originally for flexibility, but it is just a composition of other exposed functions, so perhaps it doesn't belong.

Math/NumberTheory/DirichletCharacters.hs
b-mehta: This function is not exposed but is used internally in multiple contexts, so if it's okay I'll leave this one polymorphic.

Math/NumberTheory/DirichletCharacters.hs
Bodigrim: Let's remove it then.

Math/NumberTheory/DirichletCharacters.hs
Bodigrim: OK

Math/NumberTheory/DirichletCharacters.hs
b-mehta: It would, but it would be harder to work with - even one of the tests would need to prove `Mod (d * n) d ~ 0` (which isn't in `Data.Constraint`). I think it's best to use `Divides d n` from Data.Constraint, and add the lemma `Divides a c => Divides a (b * c)`, but this would belong in a different branch as it's a bit of work.

Bodigrim: I see. How do you feel about this?
```haskell
induced :: (KnownNat d, KnownNat n) => DirichletCharacter d -> DirichletCharacter (d * n)
```

Math/NumberTheory/Moduli/Internal.hs
Bodigrim: I meant that IIRC the copyright should be yours, not mine ;)

b-mehta: Ah sure! :)

Math/NumberTheory/DirichletCharacters.hs
Bodigrim: Ah, I am terrible with names. Probably `orZeroToNum` will do, if you still prefer it to `coerce (maybe 0)`.

Math/NumberTheory/DirichletCharacters.hs
b-mehta: I also tried this; it requires `AllowAmbiguousTypes` in the module, and `TypeApplications` for the user: if `chi :: DirichletCharacter 5` then GHC can't figure out what `d` should be to make `induced chi :: DirichletCharacter 15` work, since `*` is non-injective. The user would instead need to do `induced @3 chi :: DirichletCharacter 15`, at which point we might as well use the version here, as far as I can see.

Bodigrim: Yeah, I forgot that GHC treats `*` as non-injective because of `0`.

b-mehta:
I've added `fromTable`, which allows one to construct a `DirichletCharacter` from its table of values, if a valid character exists, and also I've adjusted `iterateMaybe` to use `unfoldr` as opposed to manual list creation.

Here's a list of improvements that can be turned into issues after merge.

- `fromTable` currently requires the whole table, but only uses part of it to get the underlying character (and we verify the generated character is correct). In fact, it suffices to only supply a very small number of values to uniquely specify a character. If one specifies `chi` at a collection of points which generates `(Z/nZ)*`, this would uniquely define the character (although this might be hard to do in practice). With this in mind, we can make `fromTable` more general:
```haskell
data DirichletError = UnderSpecified | OverSpecified | InvalidValues
fromTable :: forall n. KnownNat n => Vector (OrZero RootOfUnity) -> Either DirichletError (DirichletCharacter n)
```
- `toRealFunction` uses `evalGeneral`, but it should be possible to calculate it without doing any discrete log calculations - I believe we can use `jacobiSymbol` to do the odd prime power cases for instance - which would make it more efficient
- (Easy) Conversion to/from the right sort of ArithmeticFunction (the conversion `from` ought to be partial, using `fromTable`) - this was mentioned in the original issue. If it's still desired, the conversion `from` would need an unchecked precondition that the given function is periodic. The conversion `to` should use `evalAll` since we might need to evaluate it at every prime, which cover almost all of `MultMod n`.

The following aren't directly related to characters, but they came up while working on this.

- It would be nice to have a generalised version of `chineseCoprime` which works on more than two cases
- It might be useful to export
```haskell
reduceMod :: forall a b. (KnownNat a, KnownNat b) => Mod (a*b) -> (Mod a, Mod b)
reduceMod (Mod n) = (fromIntegral n, fromIntegral n)

reduceMult :: forall a b. (KnownNat a, KnownNat b) => MultMod (a*b) -> (MultMod a, MultMod b)
reduceMult (MultMod n) = (MultMod a, MultMod b)
  where (a,b) = reduceMod n
```

Bodigrim:
Great job, thanks for your efforts and time!

----

> `fromTable` currently requires the whole table, but only uses part of it 

I'd probably go for 

```haskell
fromTable :: forall n. KnownNat n => (Mod n -> OrZero RootOfUnity) -> DirichletCharacter n
fromTable f = undefined
```

such that `fromTable` evaluates `f` on each group generator and reconstructs all the rest. And `f` may lookup into a vector or whatever it likes.

> `toRealFunction` uses `evalGeneral`, but it should be possible to calculate it without doing any discrete log calculations

That's a good point.

> Conversion to/from the right sort of `ArithmeticFunction`

I think there are enough helpers for a user willing to implement it himself, so this is of a low priority probably.

> It would be nice to have a generalised version of chineseCoprime which works on more than two cases

What exactly do you have in mind? 

> It might be useful to export `reduceMod`

This could make a nice addition to https://github.com/Bodigrim/mod

b-mehta:
> such that fromTable evaluates f on each group generator and reconstructs all the rest. And f may lookup into a vector or whatever it likes.

I don't think this makes much of a difference; besides `fromTable` still checks that the character is valid, which is done by ensuring the whole table matches.
What I had in mind was more that some values can be unspecified while still giving the character - but it would need a more advanced method than the one I used here.

> What exactly do you have in mind?

This is what I'd quickly guessed earlier, though I don't doubt something better exists
```haskell
chineseCoprimeList :: (Eq a, Ring a, Euclidean a) => [(a, a)] -> Maybe a
chineseCoprimeList xs = fmap fst $ foldM (\x y -> fmap (,snd x `times` snd y) (chineseCoprime x y)) (zero, one) xs
```

PR #184, opened 2020-01-14
Chinese remainder theorem on a list

b-mehta:
Motivation:
- It is useful to have chinese remainder theorem on more than two arguments
- `chineseRemainder` supported this but has now been deprecated, `chineseCoprimeList` restores this functionality.

I'm not sure what the right test for `chineseList` is - it's unclear to me how to generalise compatibility. Feel free to advise, or add commits to this PR.

### DIFF
--- a/Math/NumberTheory/Moduli/Chinese.hs
+++ b/Math/NumberTheory/Moduli/Chinese.hs
@@ -22,7 +22,9 @@
 module Math.NumberTheory.Moduli.Chinese
   ( -- * Safe interface
     chinese
+  , chineseList
   , chineseCoprime
+  , chineseCoprimeList
   , chineseSomeMod
   , chineseCoprimeSomeMod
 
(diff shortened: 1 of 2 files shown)


Bodigrim:
It is alarming that we need to expose `chineseList` alongside with `chinese`. Perhaps `chinese` is implemented wrong, in an anticompositional way?

----

I looked for other usages of `chinese` throughout `arithmoi`. They all follow the same pattern: 

https://github.com/Bodigrim/arithmoi/blob/54a4448a577bd2782e6ed786dae0966e6c9b8f99/Math/NumberTheory/Moduli/Equations.hs#L83-L86
https://github.com/Bodigrim/arithmoi/blob/54a4448a577bd2782e6ed786dae0966e6c9b8f99/Math/NumberTheory/Moduli/Internal.hs#L65-L67
https://github.com/Bodigrim/arithmoi/blob/54a4448a577bd2782e6ed786dae0966e6c9b8f99/Math/NumberTheory/Moduli/Sqrt.hs#L68-L71

I think the right thing here is to change `chinese` and `chineseCoprime` to return a pair:

```haskell
chineseCoprime :: (Eq a, Ring a, Euclidean a) => (a, a) -> (a, a) -> Maybe (a, a)
chinese :: forall a. (Eq a, Ring a, Euclidean a) => (a, a) -> (a, a) -> Maybe (a, a)
```

Then we can avoid adding `chineseList`, because it would be easy enough to write `foldlM chinese (0,1)`, `foldrM chinese (0,1)`, `foldr1M chinese` or whatever.

b-mehta:
Yeah that sounds sensible. I found it pretty surprising that `chineseRemainder` was deprecated without a replacement which conveniently works for lists.

Bodigrim:
Well, my original impression was that `foldlM chineseSomeMod` should do the job, but apparently `SomeMod` abtraction is a bit cumbersome.

Bodigrim:
Closing in favor of 3bc13a8de60fb059bec3874904a0eb5e3d299321.

PR #186, opened 2020-02-09
Move roots of unity to utils and ignore vim .swp files

hgsipiere:
This moves roots of unity to the utilities section.

It leaves it as an exposed module so that the tests can build.

### DIFF
--- a/Math/NumberTheory/Utils/RootsOfUnity.hs
+++ b/Math/NumberTheory/Utils/RootsOfUnity.hs
@@ -1,0 +1,59 @@
+module Math.NumberTheory.Utils.RootsOfUnity
+(  
+-- * Roots of unity
+   RootOfUnity (..)
+-- ** Conversions
+   , toRootOfUnity
+   , toComplex )
+
+where
+
+import Data.Complex                                        (Complex(..), cis)
+import Data.Semigroup                                      (Semigroup(..))
+import Data.Ratio                                          ((%), numerator, denominator)
+
+-- | A representation of <https://en.wikipedia.org/wiki/Root_of_unity roots of unity>: complex
+-- numbers \(z\) for which there is \(n\) such that \(z^n=1\).
+newtype RootOfUnity =
+  RootOfUnity { -- | Every root of unity can be expressed as \(e^{2 \pi i q}\) for some
+                -- rational \(q\) satisfying \(0 \leq q < 1\), this function extracts it.
+                fromRootOfUnity :: Rational }
+  deriving (Eq)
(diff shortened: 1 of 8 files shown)


Math/NumberTheory/Utils/RootsOfUnity.hs
```diff
@@ -0,0 +1,59 @@
+module Math.NumberTheory.Utils.RootsOfUnity
```
Bodigrim: 1. Please copy haddock header from `Math.NumberTheory.DirichletCharacters` and update it appropriately.
2. I would rather go for `Math.NumberTheory.RootsOfUnity`, especially since it needs to be exposed.

test-suite/Math/NumberTheory/DirichletCharactersTests.hs
```diff
@@ -35,6 +35,7 @@ import Math.NumberTheory.DirichletCharacters
 import qualified Math.NumberTheory.Moduli.Sqrt as J
 import Math.NumberTheory.Moduli.Class (SomeMod(..), modulo)
 import Math.NumberTheory.TestUtils (testSmallAndQuick, Positive(..))
+import Math.NumberTheory.Utils.RootsOfUnity
```
Bodigrim: Could tests for `RootsOfUnity` be possibly moved into a separate module?

Bodigrim (changes requested):

hgsipiere:
Thanks for your feed back!

I left out the copyright and maintainer, because I wasn't sure what to write.

This is small and not related to this pull request but `DirichletCharactersTests.hs` refers to `Module:       Math.NumberTheory.Moduli.DiscreteLogarithm` rather than `Math.NumberTheory.DirichletCharactersTests`.

Bodigrim:
> I left out the copyright and maintainer, because I wasn't sure what to write.

Please retain the original copyright, when moving code around. 
There may be options with regards to `Maintainer`, if the original maintainer become unreachable, but since this is not the case, I would probably retain it as well.

> This is small and not related to this pull request but DirichletCharactersTests.hs refers to...

Nice catch. It will be very much appreciated, if you fix this as well.

b-mehta:
> > This is small and not related to this pull request but DirichletCharactersTests.hs refers to...
> 
> Nice catch. It will be very much appreciated, if you fix this as well.

Seconded, thanks for this catch and fix!

Bodigrim (approved):

Bodigrim:
Thanks, @hgsipiere!

Issue #183, opened 2020-01-14
Generate divisors in range


Bodigrim:
Implement a function `divisorsInRange from to n`, which returns divisors of `n` between `from` and `to` (inclusively). The naive implementation would look this way:

```haskell
divisorsInRange :: (Ord a, UniqueFactorisation a) :: a -> a -> a -> [a]
divisorsInRange from to n = filter (\d -> from <= d && d <= to) (divisorsList n)
```

This becomes inefficient when `n` has tons of divisors out of `[from..to]` range. It would be desirable do not waste time and memory on generating divisors just to throw them away.

jyoti246:
Can you please specify the range of `n`, `from` and `to`?

Bodigrim:
`n` itself may be a huge integer, but such that its prime factorisation `factorise n` can be found quickly. Further, `0 <= from <= to <= n` and probably most of the time `from` is fairly small (like, `from = 0`) and `to` is much smaller than `n` (like `to ^ 2 < n`).

jyoti246:
I have a working code for the above mentioned issue and need help in integrating it with the ecosystem and guidance for the same. So please tell me where I can merge my code and what are the changes required and I will do the necessary changes required to be done before making a pull request.
Below is my code:
```
import System.IO
import Data.Ratio
-- import Data.IORef (newIORef, readIORef, writeIORef)

mylog :: Integer -> Integer -> Integer
mylog a b 
	| ((a `mod` b) /= 0)  = 0
	| otherwise = (1 + mylog (a `div` b) b)


getfactors :: Integer -> Integer -> [(Integer,Integer)]
getfactors 1 _ = []
getfactors n st
		| lst == []  = [(n,1)]
		| otherwise	= (((head lst), (mylog n (head lst))):getfactors ( n `div` ((head lst) ^ (mylog n (head lst) ))) ((head lst)+1))
			where lst = (take 1 $ filter (\y -> (n `mod` y) == 0) [st .. (floor (sqrt (fromIntegral n)))])
				

precalc :: Integer -> [(Integer,Integer)] -> [Integer]
precalc n [] = []
precalc n (x:[]) = [n `div` (fst x ^ snd x )] 
precalc n (x:xs) = (n `div` (fst x ^ snd x )): (precalc (n `div` ((fst x) ^ (snd x) )) xs)


func :: Integer -> Integer -> [(Integer,Integer)] -> Int -> Integer -> Integer -> [Integer] -> [Integer]
[...]

PR #190, opened 2020-03-22
Divisors in range implementation

basimkhajwal:
An implementation of #183 as an arithmetic function that keeps a sorted list and filters within the upper bound.

It doesn't currently make use of the lower bound but I don't think we could do that as an arithmetic function (correct me if I'm wrong).

The implementation could probably be made more efficient if it was as a Set rather than a List.

### DIFF
--- a/Math/NumberTheory/ArithmeticFunctions/Standard.hs
+++ b/Math/NumberTheory/ArithmeticFunctions/Standard.hs
@@ -15,6 +15,7 @@ module Math.NumberTheory.ArithmeticFunctions.Standard
   , divisors, divisorsA
   , divisorsList, divisorsListA
   , divisorsSmall, divisorsSmallA
+  , divisorsInRange, divisorsInRangeA
   , divisorCount, tau, tauA
   , sigma, sigmaA
   , totient, totientA
@@ -39,6 +40,7 @@ import qualified Data.IntSet as IS
 import Data.Maybe
 import Data.Set (Set)
 import qualified Data.Set as S
+import qualified Data.List as L
 import Data.Semigroup
 
 import Math.NumberTheory.ArithmeticFunctions.Class
@@ -96,6 +98,14 @@ divisorsHelperSmall _ 0 = IS.empty
 divisorsHelperSmall p 1 = IS.singleton p
 divisorsHelperSmall p a = IS.fromDistinctAscList $ p : p * p : map (p ^) [3 .. wordToInt a]
 {-# INLINE divisorsHelperSmall #-}
+
+-- | See `divisorsInRangeA`
+divisorsInRange :: (UniqueFactorisation n, Ord n) => n -> n -> n -> [n]
+divisorsInRange from to = runFunction (divisorsInRangeA from to)
+
+-- | The sorted list of all (positive) divisors within the range (inclusive of endpoints) of an argument, produced in lazy fashion.
+divisorsInRangeA :: (UniqueFactorisation n, Ord n) => n -> n -> ArithmeticFunction n [n]
+divisorsInRangeA from to = ArithmeticFunction (\p -> BoundedListProduct (Closed to) . divisorsListHelper (unPrime p)) (filter (>=from) . (1:) . getBoundedListProduct)
 
 -- | Synonym for 'tau'.
 --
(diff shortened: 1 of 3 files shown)


Bodigrim:
Thanks, this is a very elegant solution :)

1. I agree that we cannot make much use of the lower bound. Thus, it probably makes sense to sacrifice it altogether, making API simpler. Users can `filter (>= from)` themselves. 

```haskell
divisorsTo :: (UniqueFactorisation n, Ord n) => n -> n -> [n]
```

2. While end users will never experience it, I think it would still be nice to eliminate a possibility of combining `BoundedListProduct` with different bounds. Basically we need to parametrize `Monoid` by bound. This can be done is several ways (e. g., putting this bound on a type-level), but a low-tech solution is to simulate the `Reader` monad:

```haskell
newtype BoundedListProduct a = BoundedListProduct { getBoundedListProduct :: a -> [a] }

divisorsInRangeA from to = ArithmeticFunction (\p k -> BoundedListProduct (\bound -> takeWhile (<= bound) (divisorsListHelper (unPrime p) k))) (filter (>=from) . (1:) . ($ to) . getBoundedListProduct)
```

3. Yep, please change the return type to `Set n`. Someone interested in lazy evaluation can probably add `divisorsListInRange`, returning `[n]`, later.

basimkhajwal:
Thanks!

I've made the relevant changes and used the `Reader` monad approach that you suggested for 2 but I'm really intrigued by doing it as a type parameterisation. I've started reading about a few GHC extensions like `DataKinds` and `TypeFamilies` that seem to be on the right track but still haven't pieced together how to have a bounded constraint - I'd love some advice on where to read further!

test-suite/Math/NumberTheory/ArithmeticFunctionsTests.hs
```diff
@@ -55,13 +55,12 @@ divisorsProperty4 (NonZero n) = S.toAscList (runFunction divisorsA n) == IS.toAs
...
-  where from = a
-        to = from + b
-        n = to + c
-        expected = filter (\d -> d >= from && d <= to) $ S.toAscList (runFunction divisorsA n)
+-- | 'divisorsInRange' matches 'divisorsA' with a filter
+divisorsProperty6 :: Positive Int -> NonNegative Int -> Bool
+divisorsProperty6 (Positive a) (NonNegative b) = runFunction (divisorsInRangeA to) n == expected
+  where to = a
+        n = to + b
+        expected = S.takeWhileAntitone (<=to) (runFunction divisorsA n)
```
Bodigrim: `S.takeWhileAntitone` is a comparably new addition to `containers`, which causes [GHC 8.0.2 build](https://travis-ci.org/github/Bodigrim/arithmoi/jobs/666447515) to fail.
Please use `S.split` or `S.splitMember` instead.

Math/NumberTheory/ArithmeticFunctions/Standard.hs
```diff
@@ -340,40 +340,17 @@ instance Num a => Monoid (ListProduct a) where
...
+-- Represent as a Reader monad
+newtype BoundedSetProduct a = BoundedSetProduct { getBoundedSetProduct :: a -> Set a }
 
-instance Ord a => Semigroup (Bound a) where
-  Open      <> b         = b
-  b         <> Open      = b
-  Closed b1 <> Closed b2 = Closed (min b1 b2)
+instance (Ord a, Num a) => Semigroup (BoundedSetProduct a) where
+  BoundedSetProduct f1 <> BoundedSetProduct f2 = BoundedSetProduct f
+    where f b = s1 <> s2 <> foldMap (\n -> S.takeWhileAntitone (<=b) $ S.mapMonotonic (* n) s2) s1
```
Bodigrim: `S.takeWhileAntitone` is a comparably new addition to `containers`, which causes [GHC 8.0.2 build](https://travis-ci.org/github/Bodigrim/arithmoi/jobs/666447515) to fail.
Please use `S.split` or `S.splitMember` instead.

Math/NumberTheory/ArithmeticFunctions/Standard.hs
```diff
@@ -100,12 +100,12 @@ divisorsHelperSmall p a = IS.fromDistinctAscList $ p : p * p : map (p ^) [3 .. w
 {-# INLINE divisorsHelperSmall #-}
 
 -- | See `divisorsInRangeA`
-divisorsInRange :: (UniqueFactorisation n, Ord n) => n -> n -> n -> [n]
-divisorsInRange from to = runFunction (divisorsInRangeA from to)
+divisorsInRange :: (UniqueFactorisation n, Ord n) => n -> n -> Set n
+divisorsInRange to = runFunction (divisorsInRangeA to)
```
Bodigrim: Please do not forget to rename it to `divisorsTo` / `divisorsToA`.

Bodigrim:
> I'd love some advice on where to read further!

You can find a similar approach in https://github.com/Bodigrim/mod/blob/c2ae98f07e7eb5492990af3752d4545657efd10a/Data/Mod.hs#L60-L70

basimkhajwal:
Thanks, I've made the suggested changes but it was cleaner to implement the divisor function to be exclusive of the endpoint. I'll definitely look more into the typed approach next.

Math/NumberTheory/ArithmeticFunctions/Standard.hs
```diff
@@ -97,6 +99,22 @@ divisorsHelperSmall p 1 = IS.singleton p
 divisorsHelperSmall p a = IS.fromDistinctAscList $ p : p * p : map (p ^) [3 .. wordToInt a]
 {-# INLINE divisorsHelperSmall #-}
 
+-- | See `divisorsInRangeA`
+divisorsTo :: (UniqueFactorisation n, Ord n) => n -> n -> Set n
+divisorsTo to = runFunction (divisorsToA to)
+
+-- | The set of all (positive) divisors strictly below `to` 
+divisorsToA :: (UniqueFactorisation n, Ord n) => n -> ArithmeticFunction n (Set n)
```
Bodigrim: It is worth to add an `Integral n` constraint, because this function does not make sense for `GaussianInteger` or `EisensteinInteger`.

Bodigrim:
I'd really prefer to include `to` into the range. Otherwise it is likely to be a source of user's confusion. It should not be that bad, I believe; you probably just need a helper function along these lines:

```haskell
takeWhileLE :: Ord a => a -> Set a -> (Set a, Set a)
takeWhileLE b xs = if m then S.insert b ls else ls
  where
    (ls, m, _) = S.splitMember b xs 
```

Otherwise looks good to me.

Bodigrim:
Cool, looks good to me. Do you want to pursue that type-level approach or can I merge it as is?

basimkhajwal:
I think I'm still going to do some more reading into that before I feel comfortable giving it a go but I'll open up a new request in that case :)

Bodigrim:
Great, thanks for your efforts!

PR #191, opened 2020-03-22
Adding Divisors.hs to get all the factors of a number in a given range

### LINKED ISSUE #183: Generate divisors in range

jyoti246:
In my code the function ```func``` is basically going in a recursion to get all the factors from the prime factors we already calculated. It is first taking the first prime factor from the list from ```0``` to ```y``` number of times where y is the max count of that prime factor in the given number as  i , and for each i calling``` func``` recursively to second prime factor in the list and so on. The only optimisation will be not to go in the direction that will make the factors go out of the range and so at each point I am checking if going down the tree will make me get any factor in the range or not. If I find that I wont be getting any factor  if I proceed down that way I break at that point and wont continue in that direction.
For example lets take 4 25 48 . Here is how my tree would look like.
<pre>
                         1
                   /    /   \     \     \ 
                  1    2    4      8     16 
                     / |   / |    / |    /  
                    2  6  4 12   8 24  16   
 </pre>
The depth would be equal to the number of distinct factors and width would be equal to the number of factors in the given range.
The will do the other changes specified and commit
Fixes #183

### DIFF
--- a/Math/NumberTheory/Divisors.hs
+++ b/Math/NumberTheory/Divisors.hs
@@ -1,46 +1,124 @@
 import System.IO
 import Data.Ratio
+-- import Math.NumberTheory.Primes.factorise
+import Data.List
 
 
+
+
+headInt :: [Integer] -> Integer
+headInt (x:xs) =x
+
+
+-- mylog recTakeEachPrimetion here take two numbers a and b and gives the maximum 
+-- power of b that appears in a
 mylog :: Integer -> Integer -> Integer
 mylog a b 
 	| ((a `mod` b) /= 0)  = 0
 	| otherwise = (1 + mylog (a `div` b) b)
 
 
+
+
+-- getfactors takes two arguments the number 'n' that you want to factorise 
+-- and a number 'st' that should be kept to '2' while calling the recTakeEachPrimetion so 
+-- that prime factorization starts from 2 and goes upto sqrt(n) 
+
+
+
 getfactors :: Integer -> Integer -> [(Integer,Integer)]
 getfactors 1 _ = []
 getfactors n st
-		| lst == []  = [(n,1)]
-		| otherwise	= (((head lst), (mylog n (head lst))):getfactors ( n `div` ((head lst) ^ (mylog n (head lst) ))) ((head lst)+1))
+		| lst == [] = [(n,1)]
+		| otherwise	= 
+			(((headInt lst), (mylog n (headInt lst))):getfactors ( n `div` ((headInt lst) ^ (mylog n (headInt lst) ))) ((headInt lst)+1))
 			where lst = (take 1 $ filter (\y -> (n `mod` y) == 0) [st .. (floor (sqrt (fromIntegral n)))])
 				
 
+-- precalc if use for some precalculations that are needed to be done 
+-- to find factors of a number n. you need to pass n and the list of primefactors of n with 
+-- its count and this will produce a list of same length as the list of prime factors which contains the 
+-- residue of n after taking out all the primefactors of it which appear before the current index
+-- For example:
+-- 	precalc 150 [(2 1), (3 1), (5 2)] = [150/2 , 150/2/3 , 150/2/3/25]= [75 , 25 , 1]
+
 precalc :: Integer -> [(Integer,Integer)] -> [Integer]
 precalc n [] = []
-precalc n (x:[]) = [n `div` (fst x ^ snd x )] 
-precalc n (x:xs) = (n `div` (fst x ^ snd x )): (precalc (n `div` ((fst x) ^ (snd x) )) xs)
+precalc n ((x,y):[]) = [n `div` (x ^ y )] 
+precalc n ((x,y):xs) = (n `div` (x ^ y )): (precalc (n `div` (x ^ y)) xs)
 
 
-func :: Integer -> Integer -> [(Integer,Integer)] -> Int -> Integer -> Integer -> [Integer] -> [Integer]
-func i temp fact ind a b arr 
-	| (ind ==  (length fact)) = if (temp <= b && temp >= a) then [temp] else []
-	| ( (i <= (snd (fact!!ind))) && temp > b) = []
-	| ( (i <= (snd (fact!!ind))) && ((temp*(arr!!ind)) <= a)) = ((func (i+1) (temp * (fst (fact!!ind))) fact ind a b arr)  )
-	| (i <= (snd (fact!!ind))) = ( (func 0 (temp) fact (ind+1) a b arr) ++ (func (i+1) (temp * (fst (fact!!ind))) fact ind a b arr)  ) 
-	| otherwise = []
+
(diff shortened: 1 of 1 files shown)


Math/NumberTheory/Divisors.hs
```diff
@@ -0,0 +1,46 @@
...
+	| ( (i <= (snd (fact!!ind))) && ((temp*(arr!!ind)) <= a)) = ((func (i+1) (temp * (fst (fact!!ind))) fact ind a b arr)  )
+	| (i <= (snd (fact!!ind))) = ( (func 0 (temp) fact (ind+1) a b arr) ++ (func (i+1) (temp * (fst (fact!!ind))) fact ind a b arr)  ) 
+	| otherwise = []
+
+-- divisorsInRange :: (Ord a, UniqueFactorisation a) => a -> a -> a -> [a]
+divisorsInRange :: Integer -> Integer -> Integer -> [Integer]
+divisorsInRange from to n = func 0 1 (getfactors n 2) 0 from to (precalc n (getfactors n 2))
+	
+-- if sorted list required
+quicksort :: [Integer] -> [Integer]
```
Bodigrim: There is `Data.List.sort` in the standard library, no need to implement sorting yourself.

Math/NumberTheory/Divisors.hs
```diff
@@ -0,0 +1,46 @@
...
+mylog a b 
+	| ((a `mod` b) /= 0)  = 0
+	| otherwise = (1 + mylog (a `div` b) b)
+
+
+getfactors :: Integer -> Integer -> [(Integer,Integer)]
+getfactors 1 _ = []
+getfactors n st
+		| lst == []  = [(n,1)]
+		| otherwise	= (((head lst), (mylog n (head lst))):getfactors ( n `div` ((head lst) ^ (mylog n (head lst) ))) ((head lst)+1))
```
Bodigrim: Please avoid using `head` and resort to pattern-matching.

Math/NumberTheory/Divisors.hs
```diff
@@ -0,0 +1,46 @@
...
+			where lst = (take 1 $ filter (\y -> (n `mod` y) == 0) [st .. (floor (sqrt (fromIntegral n)))])
+				
+
+precalc :: Integer -> [(Integer,Integer)] -> [Integer]
+precalc n [] = []
+precalc n (x:[]) = [n `div` (fst x ^ snd x )] 
+precalc n (x:xs) = (n `div` (fst x ^ snd x )): (precalc (n `div` ((fst x) ^ (snd x) )) xs)
```
Bodigrim: Please avoid using `fst` / `snd` and resort to pattern-matching.

Math/NumberTheory/Divisors.hs
```diff
@@ -0,0 +1,46 @@
...
+import Data.Ratio
+
+
+mylog :: Integer -> Integer -> Integer
+mylog a b 
+	| ((a `mod` b) /= 0)  = 0
+	| otherwise = (1 + mylog (a `div` b) b)
+
+
+getfactors :: Integer -> Integer -> [(Integer,Integer)]
```
Bodigrim, line 29: I suggest using `Math.NumberTheory.Primes.factorise` to obtain a prime factorisation.

Math/NumberTheory/Divisors.hs
```diff
@@ -0,0 +1,46 @@
...
+precalc :: Integer -> [(Integer,Integer)] -> [Integer]
+precalc n [] = []
+precalc n (x:[]) = [n `div` (fst x ^ snd x )] 
+precalc n (x:xs) = (n `div` (fst x ^ snd x )): (precalc (n `div` ((fst x) ^ (snd x) )) xs)
+
+
+func :: Integer -> Integer -> [(Integer,Integer)] -> Int -> Integer -> Integer -> [Integer] -> [Integer]
+func i temp fact ind a b arr 
+	| (ind ==  (length fact)) = if (temp <= b && temp >= a) then [temp] else []
+	| ( (i <= (snd (fact!!ind))) && temp > b) = []
```
Bodigrim: You can make it more readable by putting each condition on a new line:

```haskell
| i <= snd (fact !! ind) 
, temp > b 
= []
```

Math/NumberTheory/Divisors.hs
```diff
@@ -0,0 +1,46 @@
...
+			where lst = (take 1 $ filter (\y -> (n `mod` y) == 0) [st .. (floor (sqrt (fromIntegral n)))])
+				
+
+precalc :: Integer -> [(Integer,Integer)] -> [Integer]
+precalc n [] = []
+precalc n (x:[]) = [n `div` (fst x ^ snd x )] 
+precalc n (x:xs) = (n `div` (fst x ^ snd x )): (precalc (n `div` ((fst x) ^ (snd x) )) xs)
+
+
+func :: Integer -> Integer -> [(Integer,Integer)] -> Int -> Integer -> Integer -> [Integer] -> [Integer]
```
Bodigrim: Please add a comment explaining the purpose of each argument.

jyoti246:
Sir,
 I also modified my code according to your instructions and used
"Math.NumberTheory.Primes.Factorization.Montgomery" and "Data.List" for
factorization and sorting purpose but I am facing lot of difficulties in
trying to check it. The reason I am facing difficulty in properly getting
arithmoi package on my machine with all the dependencies. One of the reason
is also that because of lock down I am not being able to access my IIT
Kharagpur campus Wifi and at some points during installations my machine is
trying to connect to campus proxy server while I am using my mobile data. I
read a lot of documentations to get all dependencies of arithmoi by now but
still getting some or the other error.
 So I am attaching my code with the changes suggested in your previous
comment but I haven't checked by running it. So it would be great  if you
can help me by suggesting some way I can setup arithmoi properly on my
machine and check the code.
 And will also make a pull request again with my previous code and  all the
comments you mentioned in the mean time.
Thanks
Jyoti Agrawal
IIT Kharagpur

Bodigrim:
Note that when replying to a GitHub notification, all attachments are removed and your message becomes a new comment in the thread. You could probably send an email to my personal address.

> So it would be great  if you can help me by suggesting some way I can setup arithmoi properly on my machine and check the code.

It should be possible to build `arithmoi` in a fairly standard way, using `cabal build` or `stack build`. Could you please forward the error messages you are stuck with? What is your environment?

Bodigrim:
Closing, since #183 has already been implemented in #190. Nevertheless, huge thanks for your efforts and exploration.

PR #194, opened 2020-04-11
Draft of CubicSymbol

federico-bongiorno:
This draft pull request is addressing Step 1 of the issue #155 (Cubic reciprocity law)
To be addressed:
- Add error messages.
- Add comments and tidy up.
- Add tests.


Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,100 @@
...
+
+instance S.Semigroup CubicSymbol where
+    (<>) = multiplication
+
+instance Show CubicSymbol where
+    show x
+        | x == Zero         = show 0
+        | x == Omega        = "ω"
+        | x == OmegaSquare  = "ω²"
+        | x == One          = show 1
```
Bodigrim: Please enable `{-# LANGUAGE LambdaCase #-}` and use pattern matching, e. g.,
```haskell
show = \case 
  Zero  -> ...
  Omega -> ...
  ...
```

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,100 @@
...
+    | symbol == OmegaSquare && remainder == 1  = Omega
+    | otherwise                                = error ""
+        where remainder = exponent `rem` 3
+
+conj :: CubicSymbol -> CubicSymbol
+conj x
+    | x == Zero || x == One      = x
+    | x == Omega                 = OmegaSquare
+    | x == OmegaSquare           = Omega
+    | otherwise                  = error ""
```
Bodigrim: Please use pattern matching instead of guards here and above.

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,100 @@
...
+    -- Check whether @prime@ satifies norm and primality conditions.
+    | primeNorm == 3                 = error "This Eisenstein number is not coprime to 3."
+    -- Return 0 if @prime@ divides @alpha@
+    | otherwise                      = toCubicSymbol residue
+
+        -- A.rem returns the remainder of the Euclidean algorithm.
+        where primeNorm = E.norm prime
+              residue = Prelude.foldr (f) 1 listOfAlphas
+              f = \x y -> A.rem (x * y) prime
+              -- The function @take@ does not accept Integer values.
```
Bodigrim: There is `genericTake` which does, if you need it, and also `replicate` (and `iterate`).

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,100 @@
...
+    | x == 0               = Zero
+    | x == 0 E.:+ 1        = Omega
+    | x == (-1) E.:+ (-1)  = OmegaSquare
+    | x == 1               = One
+    | otherwise            = error ""
+
+
+cubicSymbol :: E.EisensteinInteger -> E.EisensteinInteger -> CubicSymbol
+cubicSymbol alpha n = Prelude.foldr (g) One factorisation
+    where factorisation = P.factorise n
```
Bodigrim: It is important to avoid factorisation (which is slow, especially for Eisenstein numbers). Instead employ the cubic reciprocity law.

Bodigrim (changes requested):

Math/NumberTheory/Moduli/CubicSymbol.hs
federico-bongiorno: Done.

Math/NumberTheory/Moduli/CubicSymbol.hs
federico-bongiorno: Done.

Math/NumberTheory/Moduli/CubicSymbol.hs
federico-bongiorno: Done. But I probably won't need it.

Math/NumberTheory/Moduli/CubicSymbol.hs
federico-bongiorno: Still have to work on this one.

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,102 @@
...
+
+multiplication :: CubicSymbol -> CubicSymbol -> CubicSymbol
+multiplication Zero _ = Zero
+multiplication _ Zero = Zero
+multiplication One y = y
+multiplication x One = x
+multiplication Omega Omega = OmegaSquare
+multiplication Omega OmegaSquare = One
+multiplication OmegaSquare Omega = One
+multiplication OmegaSquare OmegaSquare = Omega
```
Bodigrim: It seems you do not use `multiplication` anywhere else. Please define `(<>)` directly.

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,102 @@
...
+
+instance S.Semigroup CubicSymbol where
+    (<>) = multiplication
+
+instance Show CubicSymbol where
+    show = \case
+        Zero         -> show 0
+        Omega        -> "ω"
+        OmegaSquare  -> "ω²"
+        One          -> show 1
```
Bodigrim: What's the point of such indirection? I'd probably just return `"1"` immediately.

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,102 @@
...
+
+
+exponentiation :: CubicSymbol -> Integer -> CubicSymbol
+exponentiation Zero _ = Zero
+exponentiation One _ = One
+exponentiation _ 0 = One
+exponentiation Omega 1 = Omega
+exponentiation Omega 2 = OmegaSquare
+exponentiation OmegaSquare 1 = OmegaSquare
+exponentiation OmegaSquare 2 = Omega
```
Bodigrim: Once you defined `instance Semigroup CubicSymbol`, you can use `Data.Semigroup.stimes` for exponentiation.

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,102 @@
...
+    OmegaSquare   -> Omega
+    One           -> One
+
+-- Converts Eisenstein integer to CubicSymbol value.
+toCubicSymbol :: E.EisensteinInteger -> CubicSymbol
+toCubicSymbol = \case
+    0               -> Zero
+    0 E.:+ 1        -> Omega
+    (-1) E.:+ (-1)  -> OmegaSquare
+    1               -> One
```
Bodigrim: It is super-important to keep pattern matching exhaustive. Please add a catch-all case, throwing a meaningful error message.

Bodigrim (changes requested):

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,161 @@
...
+
+instance S.Semigroup CubicSymbol where
+    (<>) Zero _                    = Zero
+    (<>) _ Zero                    = Zero
+    (<>) One y                     = y
+    (<>) x One                     = x
+    (<>) Omega Omega               = OmegaSquare
+    (<>) Omega OmegaSquare         = One
+    (<>) OmegaSquare Omega         = One
+    (<>) OmegaSquare OmegaSquare   = Omega
```
Bodigrim: You can rewrite it more naturally, like 
```haskell
instance S.Semigroup CubicSymbol where
    Zero <> _                    = Zero
    _ <> Zero                    = Zero
    ...
```

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,161 @@
...
+
+
+instance Show CubicSymbol where
+    show = \case
+        Zero         -> "0"
+        Omega        -> "ω"
+        OmegaSquare  -> "ω²"
+        One          -> "1"
+        otherwise    -> error ""
```
Bodigrim: What `otherwise` can happen here?

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,161 @@
...
+
+
+-- EXHAUST ALL CASES
+conj :: CubicSymbol -> CubicSymbol
+conj = \case
+    Zero          -> Zero
+    Omega         -> OmegaSquare
+    OmegaSquare   -> Omega
+    One           -> One
+    otherwise     -> error ""
```
Bodigrim: What `otherwise` can happen here?

Bodigrim (changes requested):

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,161 @@
...
+          norm = E.norm e
+          quotient = A.quot e badPowerPrime
+          badPowerPrime = badPrime ^ exponent
+          badPrime = 1 E.:+ (-1)
+
+-- CAN AVOID GUARDS?
+
+-- This function checks how many times 3 divides an integer @x@
+-- It need not check the case when @x <= 0@ as this is never called
+divideBy3 :: Integer -> Int
```
Bodigrim: Use `Math.NumberTheory.Utils.splitOff`.

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,161 @@
...
+-- Eisenstein number @factor@ such that @(1 + ω)^exponent * e = 1 + 3*factor@.
+-- Note that L.findIndex cannot return Nothing. This happens only if @e@ is
+-- not coprime with 3.
+getPrimaryDecomposition :: E.EisensteinInteger -> (Int, E.EisensteinInteger)
+getPrimaryDecomposition e = (exponent, factor)
+    where factor = unit * e
+          unit = (1 E.:+ 1)^exponent
+          Just exponent = L.findIndex (== 1) listOfRemainders 
+          listOfRemainders = Prelude.map (\x -> A.rem x 3) listOfAssociates
+          listOfAssociates = E.associates e
```
Bodigrim: You should not rely on any particular order of `associates`, its description does not provide any guarantees of such kind.

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,161 @@
...
+-- This function takes an Eisenstein number and returns its primary decomposition @(exponent, factor)@
+-- That is, given @e@ coprime with 3, it returns a unique integer (mod 6) @exponent@ and a unique
+-- Eisenstein number @factor@ such that @(1 + ω)^exponent * e = 1 + 3*factor@.
+-- Note that L.findIndex cannot return Nothing. This happens only if @e@ is
+-- not coprime with 3.
+getPrimaryDecomposition :: E.EisensteinInteger -> (Int, E.EisensteinInteger)
+getPrimaryDecomposition e = (exponent, factor)
+    where factor = unit * e
+          unit = (1 E.:+ 1)^exponent
+          Just exponent = L.findIndex (== 1) listOfRemainders
```
Bodigrim: This is a very unsafe thing, future generations may spend innumerable hours looking for a runtime error about an unexhaustive pattern. It is better to rewrite it as 

```haskell
exponent = case L.findIndex ... of 
  Just e -> e 
  Nothing -> error "..."
```

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,161 @@
...
+          (_, delta) = getPrimaryDecomposition beta
+          (j, alphaThreeFree) = factoriseBadPrime alpha
+
+
+
+-- This function takes an Eisenstein number @e@ and returns @(exponent, quotient)@
+-- where exponent is the largest integer such that @(1 - ω)^exponent@ divides @e@.
+-- @quotient@ is the quotient of @e@ by @(1 - ω)^exponent@
+factoriseBadPrime :: E.EisensteinInteger -> (Int, E.EisensteinInteger) 
+factoriseBadPrime e = (exponent, quotient)
```
Bodigrim: Use `Math.NumberTheory.Utils.splitOff`.

Bodigrim (changes requested):

Math/NumberTheory/Moduli/CubicSymbol.hs
federico-bongiorno: I am not sure I understand this issue. I added the exhaustive pattern matching and a not very meaningful error message. However the compiler gives me a warning saying that "otherwise" is redundant.

Math/NumberTheory/Moduli/CubicSymbol.hs
federico-bongiorno: Same as above.

folidota: Same as above.

Math/NumberTheory/Moduli/CubicSymbol.hs
federico-bongiorno: Same as above.

Math/NumberTheory/Moduli/CubicSymbol.hs
federico-bongiorno: I looked in `Math/NumberTheory/Quadratic.EisensteinIntegers` and they define the order in line 103. Is this something I cannot use outside of the file?

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,125 @@
...
+-- 4) Check if any of the two numbers is a zero or a unit. If it
+--    is, return their cubic residue.
+-- 5) If not, invoke cubic reciprocity by swapping @alpha@ and
+--    @beta@. Note both numbers have to be primary.
+--    Return to Step 2.
+
+-- This function takes two Eisenstein integers and returns their cubic residue character.
+-- Note that the second argument must be coprime to 3 else the algorithm returns an error.
+cubicSymbol :: EisensteinInteger -> EisensteinInteger -> CubicSymbol
+cubicSymbol alpha beta = case betaNorm `mod` 3 of
```
Bodigrim: It would probably be more straightforward to divide `beta` by `1-ω`.

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,125 @@
...
+  -- This checks whether beta is coprime to 3, i.e. divisible by @1 - ω@
+  -- In particular, it returns an error if @beta == 0@
+  0 -> error "Math.NumberTheory.Moduli.CubicSymbol: denominator is not coprime to 3."
+  _ -> cubicSymbolHelper alpha beta
+  where
+    betaNorm = norm beta
+
+cubicSymbolHelper :: EisensteinInteger -> EisensteinInteger -> CubicSymbol
+cubicSymbolHelper alpha beta = cubicReciprocity primaryRemainder primaryBeta <> newSymbol
```
Bodigrim, line 107: I suggest to inline this definition into `cubicSymbol`. These function are not large, not much point to introduce a new name and an additional indirection.

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,125 @@
...
+    -- This function outputs data such that
+    -- @(1 - ω)^jIntWord * alphaThreeFree = alpha@.
+    (jIntWord, alphaThreeFree) = splitOff (1 - ω) alpha
+
+-- This function takes an Eisenstein number and returns its primary decomposition @(powerUnit, factor)@
+-- That is, given @e@ coprime with 3, it returns a unique integer (mod 6) @powerUnit@ and a unique
+-- Eisenstein number @factor@ such that @(1 + ω)^powerUnit * e = 1 + 3*factor@.
+-- Note that L.findIndex cannot return Nothing. This happens only if @e@ is not
+-- coprime with 3. This cannot happen since @U.splitOff@ is called just before.
+getPrimaryDecomposition :: EisensteinInteger -> (Integer, EisensteinInteger)
```
Bodigrim: Could you possibly return `(CubicSymbol, EisensteinInteger)` and make corresponding changes to `extractPrimaryContributions` and `cubicHelper`? This will be more natural representation than `Integer` and `Mod 3`, I think. 

I suggest to pattern match on possible values of `remainder` and return corresponding cubic symbols directly.

Bodigrim (changes requested):

Math/NumberTheory/Moduli/CubicSymbol.hs
federico-bongiorno: Sure. The only reason why I didn't do it this way was to avoid checking multiple times if 3 divides the norm.

Math/NumberTheory/Moduli/CubicSymbol.hs
federico-bongiorno: This is probably better

Math/NumberTheory/Moduli/CubicSymbol.hs
federico-bongiorno: I will do this. Isn't this a bit slower?

Math/NumberTheory/Moduli/CubicSymbol.hs
Bodigrim: Ah, that's a good point. Then leave it as is.

Math/NumberTheory/Moduli/CubicSymbol.hs
Bodigrim: A bit, but not significantly, I suppose.

Math/NumberTheory/Moduli/CubicSymbol.hs
federico-bongiorno: I changed it but I am disappointed by the result. The code looks messier and it runs more slowly. I thought about it but I could not find any better way of implementing these changes. If you have any suggestions to improve it, let me know.

Bodigrim: That's surprising, I'll take a look next week. I should revisit `M.NT.Q.EisensteinIntegers` to check that all operations are as efficient as possible. 

In the meantime I have three suggestions:

* Instead of relying on autoderived `stimes` (which appears to be awkward) we can actually define our own, supporting zeros and negatives. This basically brings us back to your very first design, sorry for the digression.

```haskell
instance Semigroup CubicSymbol where
  (<>) = ... 
  stimes k n = case (k `mod` 3, n) of ...
```

* Could you please implement `symbolToNum :: CubicSymbol -> EisensteinInteger`?

federico-bongiorno: I did this. In the new commit, the timings went down to the previous level.

Bodigrim:
Could you please rebase your branch atop of `master`? I pushed some improvements to `M.NT.Q.EisensteinIntegers`. I switched `getPrimaryDecomposition` to return `CubicSymbol` and do not experience any significant slowdown. 

```diff
@@ -96,8 +96,7 @@ cubicReciprocity alpha beta = cubicSymbolHelper beta alpha
 extractPrimaryContributions :: EisensteinInteger -> EisensteinInteger -> (EisensteinInteger, EisensteinInteger, CubicSymbol)
 extractPrimaryContributions alpha beta = (gamma, delta, newSymbol)
   where
-    newSymbol = stimes contribution Omega
-    contribution = j*m - i*m -i*n
+    newSymbol = stimes (j * m) Omega <> stimes (- m - n) i
     m :+ n = A.quot (delta - 1) 3
     (i, gamma) = getPrimaryDecomposition alphaThreeFree
     (_, delta) = getPrimaryDecomposition beta
@@ -111,12 +110,12 @@ extractPrimaryContributions alpha beta = (gamma, delta, newSymbol)
 -- Eisenstein number @factor@ such that @(1 + ω)^powerUnit * e = 1 + 3*factor@.
 -- Note that L.findIndex cannot return Nothing. This happens only if @e@ is not
 -- coprime with 3. This cannot happen since @U.splitOff@ is called just before.
-getPrimaryDecomposition :: EisensteinInteger -> (Integer, EisensteinInteger)
+getPrimaryDecomposition :: EisensteinInteger -> (CubicSymbol, EisensteinInteger)
 -- This is the case where a common factor between @alpha@ and @beta@ is detected.
 -- In this instance @cubicReciprocity@ will return @Zero@.
 -- Strictly speaking, this is not a primary decomposition.
-getPrimaryDecomposition 0 = (0, 0)
-getPrimaryDecomposition e = (toInteger powerUnit, factor)
+getPrimaryDecomposition 0 = (Zero, 0)
+getPrimaryDecomposition e = (stimes powerUnit Omega, factor)
   where
     factor = unit * e
     unit = (1 + ω)^powerUnit
```

Math/NumberTheory/Moduli/CubicSymbol.hs
```diff
@@ -0,0 +1,132 @@
...
+  where
+    factor = unit * e
+    unit = (1 + ω)^powerUnit
+    powerUnit = fromMaybe
+      (error "Math.NumberTheory.Moduli.CubicSymbol: primary decomposition failed.")
+      findPowerUnit
+    -- Note that the units in @ids@ are ordered in the following way:
+    -- The i^th element of @ids@ is @(1 + ω)^i@ starting from i = 0@
+    -- That is the i^th unit counting anticlockwise starting with 1.
+    findPowerUnit = elemIndex inverseRemainder ids
```
Bodigrim: It seems we can do this "table lookup" more explicit, pattern-matching by all possible values of `remainder`. Pattern-matching provides an elegant way to define `symb` (equal to `stimes powerUnit Omega`), and then one can choose either `e * symbolToNum symb` or `-e * symbolToNum`, depending on which of them is equivalent to 1 modulo 3.

federico-bongiorno: As I understand it, this method is not going to work. My understanding is that you want to define `factor = e * symbolToNum symb` or `factor = -e * symbolToNum symb` depending on which one is equivalent to 1 mod 3. In general, I don't think that either of them need be equivalent to 1 mod 3. For instance, suppose that I find that `(1 + ω) * e A.rem 3 = 1`. Then its associated cubic symbol is ω. Then you see that both `e * ω` and `-e * ω` cannot have the above property. The issue is that I need to know `powerUnit` modulo 6 rather that modulo 3. One solution could be to pattern match on `remainder` and define both `factor` and `symb` at the same time. Let me know what you think.

federico-bongiorno: I have taken this approach in the latest commit. I would like to make the pattern matching more readable but the program does not compile if I use ω when I list the cases. Also, it gives me a warning for redundant pattern matching which I don't agree with.

Bodigrim:
Looks great! Please write haddock comments to the exported enitities and ensure that they are correctly rendered (`stack haddock`). You can use LaTeX (https://haskell-haddock.readthedocs.io/en/latest/markup.html#mathematics-latex).

Bodigrim:
Merged, thanks for your efforts!

PR #195, opened 2020-04-16
Cornacchia implementation

### LINKED ISSUE #192: Implement Cornacchia's algorithm
Implement Cornacchia's algorithm to solve x^2 + d*y^2 = m, where 1 ≤ d < m and d and m are coprime. 

https://projecteuclid.org/DPubS/Repository/1.0/Disseminate?view=body&id=pdf_1&handle=euclid.pja/1116442240

basimkhajwal:
Implements #192 .

### DIFF
--- a/test-suite/Math/NumberTheory/DiophantineTests.hs
+++ b/test-suite/Math/NumberTheory/DiophantineTests.hs
@@ -10,16 +10,12 @@ module Math.NumberTheory.DiophantineTests
 
 import Test.Tasty
 
-import Math.NumberTheory.Primes
 import Math.NumberTheory.Diophantine
 import Math.NumberTheory.TestUtils
 
--- Test cornacchia implementation by generating coprime (d,m) with a solution
-cornacchiaTest :: Prime Integer -> NonNegative Integer -> Positive Integer -> Bool
-cornacchiaTest p (NonNegative a) (Positive b) = all checkSoln solns && (a*d+1, b) `elem` solns
-  where d = unPrime p
-        m = (a*d+1)^2 + d*b*b
-        solns = cornacchia d m
+cornacchiaTest :: Positive Integer -> Positive Integer -> Bool
+cornacchiaTest (Positive d) (Positive a) = all checkSoln (cornacchia d m)
+  where m = d + a
         checkSoln (x, y) = x*x + d*y*y == m
 
 testSuite :: TestTree
--- a/test-suite/Math/NumberTheory/DiophantineTests.hs
+++ b/test-suite/Math/NumberTheory/DiophantineTests.hs
@@ -17,7 +17,7 @@ import Math.NumberTheory.Roots (integerSquareRoot)
 import Math.NumberTheory.TestUtils
 
 cornacchiaTest :: Positive Integer -> Positive Integer -> Bool
-cornacchiaTest (Positive d) (Positive a) = all checkSoln (cornacchia d m)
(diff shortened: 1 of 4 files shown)


Bodigrim:
Nice! While Cornacchia algorithm itself requires coprime `d` and `m`, could you possibly extend your routine to cover non-coprime `d` and `m` as well?

basimkhajwal:
Thanks! I'm not sure how I would go about doing that efficiently? 
Any pointers to any resources that I could read from for the general case?

Bodigrim:
Let `g = gcd d m` and `factorise g = [(p_1, a_1), ..., (p_n, a_n)]`. If any `a_k` is odd, there are no solutions. Otherwise, if all `a_k` are even, apply Cornacchia algorithm to `(d / g, m / g)` and scale `x`s by `sqrt g`.

basimkhajwal:
Thanks! I didn't realise that those were all the solutions. I've made the necessary changes

test-suite/Math/NumberTheory/DiophantineTests.hs
```diff
@@ -0,0 +1,22 @@
...
+  ( testSuite
+  ) where
+
+import Test.Tasty
+
+import Math.NumberTheory.Diophantine
+import Math.NumberTheory.TestUtils
+
+cornacchiaTest :: Positive Integer -> Positive Integer -> Bool
+cornacchiaTest (Positive d) (Positive a) = all checkSoln (cornacchia d m)
```
Bodigrim: Such test will pass even if we define `cornacchia _ _ = []`. Please write a reference implementation, searching for solutions using brute force, and compare the output of `cornacchia` against this reference.

Bodigrim:
Sorry, I do not have time to look inside, but this looks suspicious:
```haskell
> cornacchia 1 4
[]
```
I expected to obtain `[(0,2), (2,0)]`.

basimkhajwal:
> Sorry, I do not have time to look inside, but this looks suspicious:
> 
> ```haskell
> > cornacchia 1 4
> []
> ```
> 
> I expected to obtain `[(0,2), (2,0)]`.

Currently it outputs positive `(x,y)` as solutions but I guess I could add that in seeing as that behaviour is probably more intuitive

basimkhajwal:
It's been a while but I thought I should just finish this off now before it's left too long.

I ran some more tests against a brute force implementation and everything seems to going fine on coprime inputs, it seems however that dealing with non-coprime inputs isn't quite what I had thought would work before.

> Let `g = gcd d m` and `factorise g = [(p_1, a_1), ..., (p_n, a_n)]`. If any `a_k` is odd, there are no solutions. Otherwise, if all `a_k` are even, apply Cornacchia algorithm to `(d / g, m / g)` and scale `x`s by `sqrt g`.

I implemented this but it doesn't actually account for all solutions, as an example consider `d=2` and `m=6` which have gcd `2` which is not square but there is a solution to the overall problem being `(2,2)`.

I'm not sure if there's any way to easily consider these cases but if not I can just add coprime constraint to the inputs.

Bodigrim:
I see, now I realise that my suggestion was wrong. Sorry for pointing you in the wrong direction. Please constrain `cornacchiaPrimitive` to coprime inputs, and I'll be happy to merge.

Bodigrim:
Please take care of an hlint suggestion: https://travis-ci.org/github/Bodigrim/arithmoi/jobs/707021161#L950

Bodigrim:
Thanks, merged!

PR #198, opened 2020-05-28
Quadratic Sieve - Divide

federico-bongiorno:
In this branch, the factorisations are stored in a vector of Boolean values.

### DIFF
--- a/Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
+++ b/Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
@@ -1,60 +1,80 @@
 module Math.NumberTheory.Primes.Sieve.SmoothSieve
-  ( smoothSieve
-    smoothNumbers
+  ( smoothData,
+    smoothNumbers,
   ) where
 
-import Data.Vector (Vector, replicate, find, accumulate)
+import Data.Vector (Vector, replicate, findIndices, accumulate, map)
 import qualified Data.Vector as V
 import qualified Data.Vector.Mutable as MV
 import Control.Monad.ST
 import Control.Monad
-import Data.Maybe
 -- Finds the number of primes less than or equal to a given number
 import Math.NumberTheory.Primes.Counting.Impl
 
--- Input an integer n and b and it returns a vector whose i^th component
--- is the factorisation vector of i with respect to a factor base b as
--- i goes from 2 to n. The factor base comprises all the primes less than
--- or equal to b.
+smoothData :: Int -> Int -> (V.Vector Int, V.Vector (V.Vector Bool))
+smoothData n b = (smoothNumbers, smoothFactorisations)
+    where
+        smoothFactorisations = V.map (factorisationsArray V.!) indicesArray
+        indicesArray = V.findIndices (== 1) numbersArray
+        smoothNumbers = V.map (+2) indicesArray
+        (numbersArray, factorisationsArray) = smoothSieve n b
+
+smoothNumbers :: Int -> Int -> V.Vector Int
+smoothNumbers n b = fst (smoothData n b)
+
+-- Input an integer n and b and it returns a tuple of vectors. The first
+-- element of the tuple is the vector of the numbers 2 to n divided by
+-- all the primes in the factor base (primes less than b). The second
+-- element is a vector consisting of their factorisations modulo 2.
+-- In details, the i^th component of this vector is the factorisation
+-- vector of the i modulo 2 with respect to the factor base as
+-- i goes from 2 to n.
 -- For example, given factor base [2,3,5,7], the factorisation vector
--- of 63 is [0,2,0,1]
-smoothSieve :: Int -> Int -> V.Vector (V.Vector Int)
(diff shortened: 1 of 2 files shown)


Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
```diff
@@ -0,0 +1,80 @@
+module Math.NumberTheory.Primes.Sieve.SmoothSieve
+  ( smoothData,
+    smoothNumbers,
```
Bodigrim, line 3: Please put commas first, as in other modules. For example,
https://github.com/Bodigrim/arithmoi/blob/76b7f93653216f208a41691ad0a562644d340a9c/Math/NumberTheory/Primes.hs#L13-L22

Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
```diff
@@ -0,0 +1,80 @@
...
+  ) where
+
+import Data.Vector (Vector, replicate, findIndices, accumulate, map)
```
Bodigrim, line 6: Usual pattern is to import unqualified only `Vector`, and use `V.replicate`, `V.findIndices`, `V.accumulate`, `V.map`).

Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
```diff
@@ -0,0 +1,80 @@
...
+import qualified Data.Vector as V
+import qualified Data.Vector.Mutable as MV
+import Control.Monad.ST
+import Control.Monad
+-- Finds the number of primes less than or equal to a given number
+import Math.NumberTheory.Primes.Counting.Impl
```
Bodigrim, line 12: It is not a big deal and you certainly can do it, but if there is no specific reason, it is better to stick to publicly exposed modules such as `import Math.NumberTheory.Primes`. This is because the interface of public modules is more or less stable, while internals can be changed by other contributors without restrictions.

Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
```diff
@@ -0,0 +1,80 @@
...
+
+smoothData :: Int -> Int -> (V.Vector Int, V.Vector (V.Vector Bool))
```
Bodigrim, line 14: Even while this is a prototype, it is a good habit to use unboxed vectors whenever you see an unboxable type. Most of "small" types are unboxable: `Int`, `Word`, `Bool`, `Double`.

```haskell
import qualified Data.Vector.Unboxed as U 

smoothData :: Int -> Int -> (U.Vector Int, V.Vector (U.Vector Bool))
```

Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
```diff
@@ -0,0 +1,80 @@
...
+        primePositionM <- V.thaw primePosition
+        smoothSieveM numbersM factorisationsM primePositionM n b
+        numbersArray <- V.unsafeFreeze numbersM
+        factorisationsArray <- V.unsafeFreeze factorisationsM
+        pure (numbersArray, factorisationsArray)
+
+smoothSieveM :: MV.MVector s Int -> MV.MVector s (V.Vector Bool) -> MV.MVector s Int -> Int -> Int -> ST s ()
+smoothSieveM numbersM factorisationsM primePositionM n b = do
+    let bound = minimum [n,b]
+    forM_ [2..bound] $ \index -> do
```
Bodigrim, line 61: Use `[nextPrime 2 .. precPrime bound]` to generate a list of primes.

Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
```diff
@@ -0,0 +1,80 @@
...
+                addFactor numbersM factorisationsM primePositionM n prime divisor
+            MV.modify primePositionM (+1) 0
+
+addFactor :: MV.MVector s Int -> MV.MVector s (V.Vector Bool) -> MV.MVector s Int -> Int -> Int -> Int -> ST s ()
+addFactor numbersM factorisationsM primePositionM n prime divisor = do
+    -- This is the position storing information about prime
+    index <- MV.read primePositionM 0
+    -- There must be a better way to change the entry
+    let addComponent v = accumulate (\x _ -> not x) v (V.singleton (index, 0))
+        divide x = x `div` prime
```
Bodigrim, line 77: Usually such short definitions do not deserve a separate name. Just write ``(`div` prime)`` instead of `divide` below.

Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
Bodigrim: Also, do you need a pair of vectors or a vector of pairs is enough? `V.Vector (Int, U.Vector Bool)`.

Bodigrim (commented):
I hope you find these suggestions helpful, but in general I suggest to use `M.NT.ArithmeticFunctions.SieveBlock` and generate `V.Vector IntSet`.

PR #200, opened 2020-05-29
Quadratic Sieve - Sparse

federico-bongiorno:


Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
```diff
@@ -0,0 +1,84 @@
...
+import Data.Vector.Generic (convert)
+
+-- Finds the number of primes less than or equal to a given number
+--import Math.NumberTheory.Primes.Counting.Impl
+
+smoothData :: Int -> Int -> (U.Vector Int, V.Vector IS.IntSet)
+smoothData n b = (smoothIndices, smoothFactorisations)
+    where
+        -- Need to convert from unboxed to boxed
+        smoothFactorisations = V.map (factorisationsArray V.!) (convert indicesArray)
```
federico-bongiorno, line 23: Here I am forced to convert to boxed vectors, otherwise the output would also be unboxed

Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
```diff
@@ -0,0 +1,84 @@
...
+  , smoothNumbers
+  ) where
+
+import Control.Monad
+import Control.Monad.ST
+import Data.IntSet as IS
+import qualified Data.Vector as V
+import qualified Data.Vector.Unboxed as U
+import qualified Data.Vector.Mutable as MV
+import qualified Data.Vector.Unboxed.Mutable as UMV
```
Bodigrim, line 12: Not a big deal, but I prefer `as MU` for `Data.Vector.Unboxed.Mutable`.

Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
```diff
@@ -0,0 +1,84 @@
...
+-- To convert between boxed and unboxed
+import Data.Vector.Generic (convert)
```
Bodigrim, line 14: It is already available as `V.convert`.

Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
```diff
@@ -0,0 +1,84 @@
...
+        primePositionM <- U.thaw primePosition
+        smoothSieveM numbersM factorisationsM primePositionM n b
+        numbersArray <- U.unsafeFreeze numbersM
+        factorisationsArray <- V.unsafeFreeze factorisationsM
+        pure (numbersArray, factorisationsArray)
+
+smoothSieveM :: UMV.MVector s Int -> MV.MVector s IS.IntSet -> UMV.MVector s Int -> Int -> Int -> ST s ()
+smoothSieveM numbersM factorisationsM primePositionM n b = do
+    let bound = minimum [n,b]
+    forM_ [2..bound] $ \index -> do
```
Bodigrim, line 67: Please use [nextPrime 2 .. precPrime bound] to generate a list of primes.

Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
```diff
@@ -0,0 +1,84 @@
...
+smoothSieve n b
+    | n < 2 || b < 2 = (U.empty, V.empty)
+    | otherwise      = runST $ do
+        -- It may be more efficient to skip this step and let b
+        -- be the number of primes
+        -- Discuss whether use Int or Integer
+        let --numberOfPrimes = fromInteger (primeCount (toInteger b))
+            -- Array keeping track of divisions
+            numbers = U.generate (n - 1) (\x -> x + 2)
+            factorisations = V.replicate (n - 1) IS.empty
```
Bodigrim, line 51: It is much more convenient to allocate `n + 1` elements and index them without ±2. Eventually, you'll forger about this shift somewhere, dooming yourself to a painful debug session (and mutable vectors are not something with is easy to debug).

Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
```diff
@@ -0,0 +1,84 @@
...
+                addFactor numbersM factorisationsM primePositionM n prime divisor
+            UMV.modify primePositionM (+1) 0
+
+addFactor :: UMV.MVector s Int -> MV.MVector s IS.IntSet -> UMV.MVector s Int -> Int -> Int -> Int -> ST s ()
+addFactor numbersM factorisationsM primePositionM n prime divisor = do
+    -- This is the position storing information about prime
+    index <- UMV.read primePositionM 0
+    let addComponent set = if (IS.member index set) then (IS.delete index set) else (IS.insert index set)
+    forM_ [divisor, divisor*2..n] $ \multiple -> do
+        UMV.modify numbersM (`div` prime) (multiple - 2)
```
Bodigrim, line 83: There is a handy trick: instead of dividing every time, initialize `numbersM` with 1, then multiply by `prime` each time when `prime` divides index. And at the very last step replace products of small primes with the remaining factor ``V.imap (\i v -> i `quot` v)``. This approach improves performance, because multiplication is faster than division.

Math/NumberTheory/Primes/Sieve/SmoothSieve.hs
```diff
@@ -0,0 +1,84 @@
...
+            let powers = takeWhile (< n) (iterate (* prime) prime)
+            forM_ powers $ \divisor ->
+                addFactor numbersM factorisationsM primePositionM n prime divisor
+            UMV.modify primePositionM (+1) 0
+
+addFactor :: UMV.MVector s Int -> MV.MVector s IS.IntSet -> UMV.MVector s Int -> Int -> Int -> Int -> ST s ()
+addFactor numbersM factorisationsM primePositionM n prime divisor = do
+    -- This is the position storing information about prime
+    index <- UMV.read primePositionM 0
+    let addComponent set = if (IS.member index set) then (IS.delete index set) else (IS.insert index set)
```
Bodigrim, line 81: `IntSet` operations are relatively expensive once, especially those which insert/delete elements, because they allocate memory. It is better to perform them as seldom as possible. 

Intuitively I would rather to detect exact powers first, even while it involves divisions. Then if a power is even, you don't need to modify `IntSet` at all. And if it is odd, insert a new element just once. 

This is probably worth to benchmark.

PR #208, opened 2020-06-23
Binary Linear Algebra

federico-bongiorno:
An implementation of sparse binary vectors and matrices with linear algebra solver.


Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -73,8 +74,10 @@ linear matrix@(SBMatrix m) = minPoly
   where
     z = SBVector (S.fromList $ randomSublist (I.keys m) (mkStdGen (fromIntegral (unsafePerformIO getMonotonicTimeNSec))))
     randomSequence = generateData matrix z
-    errorPoly = toF2Poly $ U.update (U.replicate (2 * (size matrix) + 1) (Bit False)) (U.singleton (2 * (size matrix), Bit True))
-    minPoly = snd $ gcdExt randomSequence errorPoly
+    dim = size matrix
+    -- Is this the best way to generate @errorPoly@?
```
Bodigrim: Use ``fromInteger (1 `shiftL` (2*dim)) :: F2Poly`` or something similar.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,112 @@
...
+    -- This assumes rows are indexed in the same way as columns are.
+    primes = [1..(size matrix)]
+
+berlekampMassey :: Int -> F2Poly -> F2Poly -> F2Poly
+berlekampMassey dim = go 1 0
+  where
+    -- Is there a better way to implement recursion in this situation?
+    go :: F2Poly -> F2Poly -> F2Poly ->F2Poly -> F2Poly
+    go oneBefore twoBefore a b
+      -- Better way to compute degree
```
Bodigrim: `U.length` is actually good performance-wise.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,112 @@
...
+    -- This should be replaced by factorBase = [nextPrime 2..precPrime b]
+    -- This assumes rows are indexed in the same way as columns are.
+    primes = [1..(size matrix)]
+
+berlekampMassey :: Int -> F2Poly -> F2Poly -> F2Poly
+berlekampMassey dim = go 1 0
+  where
+    -- Is there a better way to implement recursion in this situation?
+    go :: F2Poly -> F2Poly -> F2Poly ->F2Poly -> F2Poly
+    go oneBefore twoBefore a b
```
Bodigrim, line 137: It would be nice to state in comments an invariant relation between `oneBefore`, `twoBefore`, `a` and `b`.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,112 @@
...
+      -- Better way to compute degree
+      | U.length (unF2Poly b) <= dim + 1 = oneBefore
+      | otherwise                      = go (twoBefore - oneBefore * q) oneBefore b r
```
MathiasBartl: Formating

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,112 @@
...
+instance Semigroup SBMatrix where
+  (<>) = (*)
+
+size :: SBMatrix -> Int
+size (SBMatrix m) = I.size m
+
+randomSublist :: [Int] -> StdGen -> [Int]
+randomSublist list gen = fmap fst $ filter snd $ zip list (randoms gen)
+
+-- linearSolve :: SBMatrix -> SBVector
```
MathiasBartl: Why are you commiting with unused code?

Bodigrim: Just to clarify the process: I asked @folidota to dump a snapshot of his work every day, so sometimes he has to comment out unfinished work just to get everything else compiling. I'm fine with it as long as this PR is marked as a draft.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,257 @@
...
+
+-- Vectors form a group (in particular a monoid) under addition.
+instance Semigroup SBVector where
+  SBVector v1 <> SBVector v2 = SBVector ((v1 S.\\ v2) <> (v2 S.\\ v1))
+
+instance Monoid SBVector where
+  mempty = SBVector mempty
+
+dot :: SBVector -> SBVector -> Bit
+dot (SBVector v1) (SBVector v2) = if even (S.size (v1 `S.intersection` v2)) then Bit True else Bit False
```
Bodigrim: Nice! You can actually simplify it to `Bit (even (S.size ...))`

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,257 @@
...
+findSolution (_ : xs) matrix vector = if result == mempty then vector else findSolution xs matrix result
+  where
+    result = matrix `mult` vector
+
+-- This routine takes a polynomial p, a matrix A and a vector w and
+-- returns p(A)w. It assumes the first coefficient of p is non zero,
+-- in particular that it is non empty. This makes the implementation
+-- easier as there is no need to write the identity matrix.
+evaluate :: SBMatrix -> SBVector -> [Bit] -> SBVector
+evaluate matrix w = foldr (\coeff acc -> (matrix `mult` acc) <> (if coeff == Bit True then w else mempty)) mempty
```
Bodigrim: `if unBit coeff ...`

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,257 @@
...
+    potentialSolution = findSolution singularities matrix almostZeroVector
+    almostZeroVector = evaluate matrix z reducedMinPoly
+    (singularities, reducedMinPoly) = L.break (== Bit True) (U.toList $ unF2Poly potentialMinPoly)
+    -- lowest common multiple of previousPoly and candidateMinPoly CHECK
+    potentialMinPoly = lcm previousPoly candidateMinPoly
+    candidateMinPoly = findCandidatePoly matrix z
+
+findSolution :: [Bit] -> SBMatrix -> SBVector -> SBVector
+findSolution [] _ _ = mempty
+findSolution (_ : xs) matrix vector = if result == mempty then vector else findSolution xs matrix result
```
Bodigrim: It is better to use a guard here.
```haskell
findSolution (_ : xs) matrix vector 
  | result == mempty = vector
  | otherwise = findSolution xs matrix result
```

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,257 @@
...
+  | null singularities          = trace ("Bad z.") $ linearSolve matrix
+  -- In this case potentialMinPoly is not a good approximation to minPoly
+  -- but vec is a good random vector.
+  | potentialSolution == mempty = trace ("Bad x.") $ linearSolveHelper potentialMinPoly matrix z
+  -- This is a good solution.
+  | otherwise                   = potentialSolution
+  where
+    potentialSolution = findSolution singularities matrix almostZeroVector
+    almostZeroVector = evaluate matrix z reducedMinPoly
+    (singularities, reducedMinPoly) = L.break (== Bit True) (U.toList $ unF2Poly potentialMinPoly)
```
Bodigrim: Try `U.break`.
http://hackage.haskell.org/package/vector-0.12.0.3/docs/Data-Vector-Unboxed.html#v:break

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,257 @@
...
+
+-- Informal way of testing large matrices
+-- Input number of columns of matrix and sparsity coefficient
+testLinearSolver :: Int -> Double -> [Int]
+testLinearSolver dim sparsity = S.toList . set $ linearSolve matrix
+  where
+    numberOfRows = dim - 1
+    randomColumn :: StdGen -> [Int]
+    randomColumn gen = fmap fst $ filter (\(_, rDouble) -> rDouble < sparsity) $ zip [101..(100 + numberOfRows)] (randomRs (0, 1) gen)
+    listMatrix = map (\index -> (index, SBVector (S.fromList $ randomColumn (mkStdGen (fromIntegral (unsafePerformIO getMonotonicTimeNSec)))))) [1..dim]
```
Bodigrim: Shouldn't it be `[0..dim-1]`?

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -30,6 +31,8 @@ instance Semigroup SBVector where
 instance Monoid SBVector where
   mempty = SBVector mempty
 
+dot :: SBVector -> SBVector -> Bit
+dot (SBVector v1) (SBVector v2) = if even (S.size (v1 `S.intersection` v2)) then Bit True else Bit False
 -- Sparse Binary Matrix
 newtype SBMatrix = SBMatrix (I.IntMap SBVector) deriving (Show)
```
Bodigrim: Since you never insert or remove rows, I think it would work faster with `Data.Vector.Vector` instead of `I.IntMap`.

Bodigrim (commented):
Please rebase your branch atop of master.

app/LinearAlgebra.hs
```diff
@@ -0,0 +1,4 @@
+import Math.NumberTheory.Primes.Factorisation.LinearAlgebra
+
+main :: IO ()
+main = print $ testLinearSolver 100 0.1
```
Bodigrim: I think you vastly overestimate sparsity of matrices, relevant to quadratic sieve here. This may be the reason of your algorithm being slow. 

For instance, `arithmoi-quadratic-sieve` generates 5621 equations over 1489 primes, where only 39701 elements are non-zero. This correponds to sparsity 0.005.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,135 @@
...
+instance Monoid SBVector where
+  mempty = SBVector mempty
+
+dot :: SBVector -> SBVector -> Bit
+dot (SBVector v1) (SBVector v2) = Bit (even (S.size (v1 `S.intersection` v2)))
+-- Sparse Binary Matrix
+newtype SBMatrix = SBMatrix (V.Vector SBVector) deriving (Show)
+
+mult :: SBMatrix -> SBVector -> SBVector
+mult (SBMatrix matrix) (SBVector vector) = foldMap (matrix V.!) (S.toList vector)
```
Bodigrim: `foldMap` will allocate a lot of intermediate sets in memory. Another approach is to flip bits in a bit vector and then convert it back to set:

```haskell
mult :: SBMatrix -> SBVector -> SBVector
mult (SBMatrix matrix) (SBVector vector) = runST $ do
  -- it would be better to cache maximal length somewhere
  let len = getMax (fromJust (foldMap (fmap (Max . fst) . S.maxView . set) matrix)) + 1
  vs <- MU.new len
  traverse_ (traverse_ (unsafeFlipBit vs) . S.toList . set . (matrix V.!)) (S.toList vector)
  ws <- U.unsafeFreeze vs
  pure $ SBVector $ S.fromDistinctAscList $ listBits ws
```

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,135 @@
...
+import Data.Semigroup()
+import System.Random
+import System.IO.Unsafe
+import GHC.Clock
+import Data.Bit
+import Data.Bits
+
+-- Sparse Binary Vector
+newtype SBVector = SBVector
+  { set :: S.IntSet
```
Bodigrim: An established pattern is to call newtype's fields with a prefix `un`, so `unSBVector` in this case.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,135 @@
...
+instance Semigroup SBVector where
+  SBVector v1 <> SBVector v2 = SBVector ((v1 S.\\ v2) <> (v2 S.\\ v1))
+
+instance Monoid SBVector where
+  mempty = SBVector mempty
+
+dot :: SBVector -> SBVector -> Bit
+dot (SBVector v1) (SBVector v2) = Bit (even (S.size (v1 `S.intersection` v2)))
+-- Sparse Binary Matrix
+newtype SBMatrix = SBMatrix (V.Vector SBVector) deriving (Show)
```
Bodigrim: Please define a getter `unSBMatrix`.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,135 @@
...
+mult :: SBMatrix -> SBVector -> SBVector
+mult (SBMatrix matrix) (SBVector vector) = foldMap (matrix V.!) (S.toList vector)
+
+size :: SBMatrix -> Int
+size (SBMatrix m) = V.length m
+
+linearSolve :: SBMatrix -> SBVector
+linearSolve matrix = linearSolveHelper 1 matrix randomVectors 0 0
+  where
+    -- Make sure random vectors are not empty. 0.05 is the sparsity of random vectors
```
Bodigrim: 0.05 or 0.5?

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
federico-bongiorno: Would that be very similar to implement dense matrices?

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,135 @@
...
+  where
+    sol = linearSolve mat
+    mat = SBMatrix (V.fromList listOfColumns)
+    -- -2 is arbitrary. It means that the number of rows is at most one less than
+    -- the number of columns
+    listOfColumns = L.take dim $ getRandomVectors [0..(dim - 2)] sparsity $ mkStdGen $ fromIntegral $ unsafePerformIO getMonotonicTimeNSec
+
+
+-- 1) Investigate rare cases where code does not terminates.
+-- 2) Implement dense matrices rather than sparse.
```
Bodigrim: It could be interesting to use dense matrices to speed up multiplication, but I don't quite understand how to fit it into memory. In our toy example we have 5621 equations over 1489 primes, which is 8М bits and 1M bytes. This is easily doable. But AFAIU it is readily expected to have 1000 times more equations, isn't it? What number of equations is quoted in references?

Bodigrim (commented):
Could you please check that time grows proportionally to n^2, benchmarking matrices of growing sizes? If this is the case, then having a large constant factor is eventually manageable, but if it is not so, we are in a bigger trouble.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
Bodigrim: See my comment below: https://github.com/Bodigrim/arithmoi/pull/208#discussion_r449803481

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
federico-bongiorno: I deleted the getter all together. I don't use it anywhere

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
federico-bongiorno: I am quoting from the slides.
Researchers broke a 768 bits RSA key using NFS with Wiedemann's algorithm. The input binary matrix had 200 millions of rows and 150 non zeros elements by rows. It took 98 days of computations on a cluster of 576 cores.

Bodigrim: In this case sparse representation would take 200*10^6 * 150 * 8 = 240 Gb, which is feasible for a cluster of their size. But dense representation would take something like 1250000 Gb, which is far beyond current capabilities.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
Bodigrim: `Max` / `getMax` are from `Data.Semigroup`.
`traverse_` is from `Data.Foldable`.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,220 @@
...
+--     rows :: U.Vector Int
+--   }
+
+-- instance Semigroup DBVector where
+--   DBVector v1 <> DBVector v2 = DBVector $ zipBits xor v1 v2
+
+instance KnownNat k => Semigroup (DBVector k) where
+  DBVector v1 <> DBVector v2 = DBVector v
+    where
+      v = fromJust . SU.toSized $ zipBits xor (SU.fromSized v1) (SU.fromSized v2)
```
Bodigrim: I found a bit cleaner way to do it:
```haskell
      v = SU.withVectorUnsafe (zipBits xor (SU.fromSized v1)) v2
```

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,220 @@
...
+        mat = SBMatrix (V.fromList listOfColumns)
+        -- -2 is arbitrary. It means that the number of rows is at most one less than
+        -- the number of columns
+        listOfColumns = L.take dim $ getRandomSBVectors (dim - 1) density $ mkStdGen $ fromIntegral $ unsafePerformIO getMonotonicTimeNSec
+
+-- Infinite lists of random SBVectors. Only used for testing.
+getRandomSBVectors :: Int -> Double -> StdGen -> [SBVector]
+getRandomSBVectors numberOfRows density gen = go randomEntries
+  where
+    randomEntries = zip (cycle [0..(numberOfRows - 1)]) (randomRs (0, 1) gen)
```
Bodigrim: Running with `+RTS -h -L300` shows a spike of allocations in `randomEntries`. This is a red flag, because ideally it should not allocate at all, being fused with vector creation. Forcing `newVector` before appending it to a list fixes the issue:

```haskell
    go list = newVector `seq` newVector : go backOfList
```

For `dim = 1000`:
```
  21,980,228,616 bytes allocated in the heap
     196,614,656 bytes copied during GC
       1,156,552 bytes maximum residency (162 sample(s))
          38,368 bytes maximum slop
               1 MB total memory in use (0 MB lost due to fragmentation)
```

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,220 @@
...
+  { unSBVector :: U.Vector Int
+  } deriving (Show)
+
+-- Dense Binary Vector
+-- newtype DBVector = DBVector
+--   { unDBVector :: U.Vector Bit
+--   } deriving (Show, Eq)
+
+newtype DBVector (k :: Nat) = DBVector
+  { unDBVector :: SU.Vector k Bit
```
Bodigrim: We can actually push this approach further (and it is a good exercise), encoding in types that the matrix is a square one. 

```haskell
import GHC.TypeNats hiding (Mod)
import Data.Mod.Word
import Unsafe.Coerce

-- Sparse Binary Vector
newtype SBVector (k :: Nat) = SBVector { unSBVector :: U.Vector (Mod k) }
  deriving (Show)

-- Dense Binary Vector
newtype DBVector (k :: Nat) = DBVector { unDBVector :: SU.Vector k Bit }
  deriving (Eq, Show)

-- Sparse Binary Matrix
newtype SBMatrix (k :: Nat) = SBMatrix { unSBMatrix :: SV.Vector k (SBVector k) }
  deriving (Show)

listBits' :: KnownNat k => DBVector k -> [Mod k]
listBits' = unsafeCoerce listBits

flipBit' :: KnownNat k => MUS.MVector k s Bit -> Mod k -> ST s ()
flipBit' = unsafeCoerce (unsafeFlipBit :: MU.MVector s Bit -> Int -> ST s ())

index' :: KnownNat k => SBMatrix k -> Mod k -> SBVector k
index' = unsafeCoerce SV.index
```

(There is nothing `unsafe` in these coercions, I'm just lazy to jump through all `newtype` hoops)

Put these extra deps into `stack.yaml`:

```yaml
extra-deps:
- random-1.2.0
- smallcheck-1.2.0
- splitmix-0.1
- github: Bodigrim/mod
  commit: 64e3b10
```

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
Bodigrim: As discussed elsewhere, `seq` improves matters only for profile build. It seems that tuples are another source of thunks here, which can be easily eliminated:
```haskell
getRandomSBVectors :: Int -> Double -> StdGen -> [SBVector]
getRandomSBVectors numberOfRows density gen = go randomEntries
  where
    randomEntries = map (< density) $ randomRs (0, 1) gen
    go :: [Bool] -> [SBVector]
    go list = newVector `seq` (newVector : go backOfList)
      where
        newVector = SBVector (U.fromList listOfEntries)
        listOfEntries = map fst $ filter snd $ zip [0..] frontOfList
        (frontOfList, backOfList) = L.splitAt numberOfRows list
```
For `testLinearSolver 10000 0.01` it results in
```
  39,942,493,280 bytes allocated in the heap
     375,558,312 bytes copied during GC
      28,689,968 bytes maximum residency (33 sample(s))
       3,348,944 bytes maximum slop
              27 MB total memory in use (0 MB lost due to fragmentation)
```

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,226 @@
...
+
+
+--------------------------------------------------------------------------------
+
+testGauss :: Int -> Double -> Bool
+testGauss dim density = sol /= S.empty
+  where
+    sol = gaussianElimination mat
+    mat = map (\(a, b) -> (S.singleton a, b)) $ zip [0..] listOfColumns
+    listOfColumns = L.take dim $ getRandomIntSets (dim - 2) density $ mkStdGen $ fromIntegral $ unsafePerformIO getMonotonicTimeNSec
```
Bodigrim: Unfortunately, `GHC.Clock` appeared only in `base-4.11` and so requires GHC >= 8.4. I suggest we use [`System.CPUTime.getCPUTime`](http://hackage.haskell.org/package/base-4.10.0.0/docs/System-CPUTime.html) as a source of entropy.

Bodigrim:
There is a couple of hlint suggestions: https://travis-ci.org/github/Bodigrim/arithmoi/jobs/709224118#L1018

Bodigrim:
And one more thing to make Travis build happy: add `mod-0.1.2.0` to `extra-deps` in `stack-8.8.3.yaml`.

Bodigrim:
Continuing our discussion elsewhere. The current submission does not eat all the RAM. Yes, allocating 5 millions of `(Integer, SignedPrimeIntSet)` takes quite a lot of space, more than 500 Mb, but it works fine (and returns correct answer) with 2 Gb limit. 

While you generate `sievingInterval` of length 5000000 further computations operate over much smaller subset of smooth numbers, about 30000 elements. Instead of sieving 5000000 elements in one go (which causes a spike in memory allocations and does not scale), you can take first 10^5 values, sieve out non-smooth ones, than take next 10^5 elements and sieve them, etc., concatenating smooth vectors together.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,156 @@
...
+index' :: KnownNat k => SBMatrix k -> Mod k -> SBVector k
+index' = unsafeCoerce SV.index
+
+dot :: KnownNat k => DBVector k -> DBVector k -> Bit
+dot (DBVector v1) (DBVector v2) = Bit $ odd . countBits $ zipBits (.&.) (SU.fromSized v1) (SU.fromSized v2)
+
+mult :: KnownNat k => SBMatrix k -> DBVector k -> DBVector k
+mult matrix vector = runST $ do
+  vs <- SMU.new
+  traverse_ (traverse_ (flipBit' vs) . U.toList . unSBVector . (matrix `index'`)) $ listBits' vector
```
Bodigrim: Please use `U.mapM_` instead of `traverse_ ... . U.toList`.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,156 @@
...
+  ws <- SU.unsafeFreeze vs
+  pure $ DBVector ws
+
+size :: KnownNat k => SBMatrix k -> Int
+size (SBMatrix m) = SV.length m
```
Bodigrim: This function feels redundant to me, you can use `natVal` to demote size of the matrix from the type level to the term level.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
Bodigrim: Also `(== Bit True)` is just `unBit`.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,156 @@
...
+
+-- The input is a matrix B and a random vector z
+generateData :: KnownNat k => SBMatrix k -> DBVector k -> DBVector k -> F2Poly
+generateData matrix z x = toF2Poly $ U.fromList $ reverse $ map (x `dot`) matrixPowers
+  where
+    matrixPowers = L.take (2 * size matrix) $ L.iterate (matrix `mult`) z
+
+-- Infinite lists of random DBVectors.
+getRandomDBVectors :: KnownNat k => Int -> Double -> StdGen -> [DBVector k]
+getRandomDBVectors numberOfColumns density gen = go $ randomRs (0, 1) gen
```
Bodigrim: It seems redundant to pass `numberOfColumns` as a separate argument, because you already have it as `k` on the type level. One can use `natVal` to demote it as a term.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -129,40 +141,74 @@ appearsOnlyOnce = fst . L.foldl' go (mempty, mempty)
...
+                    primeTranslation :: [Int]
+                    primeTranslation = binarySearch (PS.toAscList (primeSet x)) $ indexPrimes columns
+
+linearSolve' :: SomeKnown SBMatrix -> SomeKnown DBVector
+linearSolve' (SomeKnown m) = SomeKnown (linearSolve m)
+
+indexPrimes :: [SignedPrimeIntSet] -> U.Vector (Prime Int)
+indexPrimes list = U.fromList . PS.toAscList $ go list
+  where
+    go :: [SignedPrimeIntSet] -> PS.PrimeIntSet
```
Bodigrim: Isn't `go` equivalent to `foldMap primeSet`?

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -129,40 +141,74 @@ appearsOnlyOnce = fst . L.foldl' go (mempty, mempty)
...
+binarySearch :: (Eq a, Ord a, U.Unbox a) => [a] -> U.Vector a -> [Int]
+binarySearch list v = go 0 (len - 1) list v
+  where
+    len = U.length v
+    go :: (Eq a, Ord a, U.Unbox a) => Int -> Int -> [a] -> U.Vector a -> [Int]
+    go _ _ [] _ = []
+    go lowerIndex upperIndex allItems@(item : otherItems) vector = case item `compare` entry of
+      LT -> go lowerIndex (currentIndex - 1) allItems vector
+      -- @(currentIndex + 1)@ makes sure no prime number index is 0
+      -- 0 is reserved for negative numbers
```
Bodigrim: This is really unexpected semantics for a function, named `binarySearch`. If users of this function need indices starting from 1, they can apply `map (+1)` themselves.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -129,40 +141,74 @@ appearsOnlyOnce = fst . L.foldl' go (mempty, mempty)
...
+      EQ -> (currentIndex + 1) : go (currentIndex + 1) len otherItems vector
```
Bodigrim: Shouldn't it be `len - 1`, similar to above?

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
federico-bongiorno: I added a comment to explain it better. Suppose I have Prime 2, Prime 3 and Prime 5. I want to send them respectively to 1,2,3. I use 0 to when the number is negative (its sign is true).

federico-bongiorno: To make it clearer `30` would be given by `[1,2,3]` whereas `-30` by `[0,1,2,3]`.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,157 @@
...
+index' :: KnownNat k => SBMatrix k -> Mod k -> SBVector k
+index' = unsafeCoerce SV.index
+
+-- | It takes a random seed and a square singular matrix and it returns an
+-- elemnent of its kernel. It does not check if the matrix is singular.
+linearSolve :: KnownNat k => StdGen -> SBMatrix k -> DBVector k
+linearSolve seed matrix = linearSolveHelper 1 matrix randomVectors 1
+  where
+    -- The floating point number is the density of the random vectors.
+    randomVectors = getRandomDBVectors 0.1 seed
```
Bodigrim: I really do not see what good may come from generating random vectors with such bias. Let's use 0.5.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,157 @@
...
+    errorPoly = fromInteger (1 `shiftL` (2 * dim)) :: F2Poly
+    dim = naturalToInt $ natVal matrix
+
+-- This routine takes a polynomial @p@, a @matrix@ and a vector @w@ and
+-- returns @p(A)w@.
+evaluate :: KnownNat k => SBMatrix k -> DBVector k -> [Bit] -> DBVector k
+evaluate matrix w = foldr (\coeff acc -> (matrix `mult` acc) <> (if unBit coeff then w else mempty)) mempty
+
+-- Tries to infer a solution. If it does not succeed, it returns an empty solution.
+findSolution :: KnownNat k => [Bit] -> SBMatrix k -> DBVector k -> DBVector k
```
Bodigrim: This function never uses values of its first arguments, only its length. That said, let's pass the length of singularities prefix as `Int` instead of `Bit`.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,157 @@
...
+findCandidatePoly :: KnownNat k => SBMatrix k -> DBVector k -> DBVector k -> F2Poly
+findCandidatePoly matrix z x = berlekampMassey dim errorPoly randomSequence
+  where
+    randomSequence = generateData matrix z x
+    errorPoly = fromInteger (1 `shiftL` (2 * dim)) :: F2Poly
+    dim = naturalToInt $ natVal matrix
+
+-- This routine takes a polynomial @p@, a @matrix@ and a vector @w@ and
+-- returns @p(A)w@.
+evaluate :: KnownNat k => SBMatrix k -> DBVector k -> [Bit] -> DBVector k
```
Bodigrim: Please use `F2Poly` instead of `[Bit]` as an argument.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,157 @@
...
+
+-- This routine finds the generating polynomial of the random sequence computed
+-- in @generateData@.
+berlekampMassey :: Int -> F2Poly -> F2Poly -> F2Poly
+berlekampMassey dim = go 1 0
+  where
+    go :: F2Poly -> F2Poly -> F2Poly ->F2Poly -> F2Poly
+    go oneBefore twoBefore a b
+      | U.length (unF2Poly b) <= dim = oneBefore
+      -- Updated value is given by @twoBefore - oneBefore * q@
```
Bodigrim: This comment literally repeats the next line of code and IMO is redundant.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,157 @@
...
+  pure $ DBVector ws
+
+listBits' :: KnownNat k => DBVector k -> [Mod k]
+listBits' = unsafeCoerce listBits
+
+flipBit' :: KnownNat k => SMU.MVector k s Bit -> Mod k -> ST s ()
+flipBit' = unsafeCoerce (unsafeFlipBit :: MU.MVector s Bit -> Int -> ST s ())
+
+index' :: KnownNat k => SBMatrix k -> Mod k -> SBVector k
+index' = unsafeCoerce SV.index
```
Bodigrim: As we are preparing for a merge, let's clean up `unsafeCoerce`. Unfortunately you'd have to change `Mod k` to just `Int` and import 

```haskell
import qualified Data.Vector.Generic.Sized.Internal as GSI
import qualified Data.Vector.Generic.Mutable.Sized.Internal as GMSI
```

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,157 @@
...
+
+-- | It takes a random seed and a square singular matrix and it returns an
+-- elemnent of its kernel. It does not check if the matrix is singular.
+linearSolve :: KnownNat k => StdGen -> SBMatrix k -> DBVector k
```
Bodigrim: Let's separate concerns. The caller of `linearSolve` does not neccessarily know which method of random generation is employed by `LinearAlgebra` and really should not care. Please refactor `linearSolve` to consume `seed :: Int` and construct `StdGen` internally.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -0,0 +1,157 @@
...
+      where
+        newVector = DBVector $ fromJust $ SU.fromList $ map (\d -> Bit (d < density)) frontOfList
+        (frontOfList, backOfList) = L.splitAt numberOfColumns list
+
+-- The input is a matrix @B@ and random vectors @z@ and @x@. It returns the
+-- sequence @x A ^ k z@ as k runs from 0 to the size of the matrix.
+generateData :: KnownNat k => SBMatrix k -> DBVector k -> DBVector k -> F2Poly
+generateData matrix z x = toF2Poly $ U.fromList $ reverse $ map (x `dot`) matrixPowers
+  where
+    matrixPowers = take (((*2) . naturalToInt . natVal) matrix) $ L.iterate (matrix `mult`) z
```
Bodigrim: It makes sense to factor out a helper `intVal = naturalToInt . natVal` and use it in three places.

Bodigrim:
Yay! Thanks!

PR #210, opened 2020-08-11
Log sieve

federico-bongiorno:
Sieving after taking logarithms


Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -14,173 +14,222 @@ import Data.Semigroup
...
+-- | Given an odd positive composite Integer @n@ and Int parameters @b@, @t@ and
+-- @k@, the Quadratic Sieve attempts to output @factor@, a factor of @n@. The
 -- parameter @b@ controls the size of the factor base. This consists of all
 -- the relevant primes which are less than or equal to @b@. The parameter @t@
--- controls the length of the sieving interval.
-quadraticSieve :: Integer -> Int -> Int -> Integer
-quadraticSieve n b t = findFactor n $ findSquares n b t
+-- controls the length of the sieving interval. The parameter @k@ controls the
+-- maximum number of blocks for the sieve to go through (@2 ^ (k - 1)@).
+quadraticSieve :: Integer -> Int -> Int -> Int -> Integer
```
Bodigrim: Having three `Int`s is error-prone: one can easily swap them. Let's introduce
```haskell
data QuadraticSieveConfig = QuadraticSieveConfig 
  { qscFactorBase :: Int 
  , qscSievingInterval :: Int 
  , qscNumberOfBlocks :: Int 
  }
```
(you can come up with better names, these are just for example).

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -14,173 +14,222 @@ import Data.Semigroup
...
-    squareRoot = integerSquareRoot n
-    mappingFunction j = j * j - n
-    startingPoint = squareRoot - intToInteger t `div` 2
+    factorBase = [nextPrime 2..precPrime t]
+    -- Make sure rootOfA is an Int
+    rootOfA = ((2 * fromInteger n) ** (1 / (4 * fromIntegral k))) / (fromIntegral m ** (1 / (2 * fromIntegral k))) :: Double
+    listOfFactors = trace ("Root of a: " ++ show rootOfA) $ generatePrimes n rootOfA k
+    decompositionOfA = zip (map (fromJust . toPrimeIntegral) listOfFactors) (repeat 2)
+    a = foldr (\(p, i) acc -> acc * unPrime p ^ i) 1 decompositionOfA
+    valuesOfB = trace ("Ratio: " ++ show ((fromInteger (a * fromIntegral m) :: Double) / (fromInteger (integerSquareRoot (2*n))))) $ filter (<= a `div` 2) $ sqrtsModFactorisation n decompositionOfA
```
Bodigrim: Split the line into two - it is too long.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -14,173 +14,222 @@ import Data.Semigroup
...
-    factorBase = [nextPrime 2..precPrime b]
-    squareRoot = integerSquareRoot n
-    mappingFunction j = j * j - n
-    startingPoint = squareRoot - intToInteger t `div` 2
+    factorBase = [nextPrime 2..precPrime t]
+    -- Make sure rootOfA is an Int
+    rootOfA = ((2 * fromInteger n) ** (1 / (4 * fromIntegral k))) / (fromIntegral m ** (1 / (2 * fromIntegral k))) :: Double
+    listOfFactors = trace ("Root of a: " ++ show rootOfA) $ generatePrimes n rootOfA k
+    decompositionOfA = zip (map (fromJust . toPrimeIntegral) listOfFactors) (repeat 2)
+    a = foldr (\(p, i) acc -> acc * unPrime p ^ i) 1 decompositionOfA
```
Bodigrim: Isn't it [`factorBack`](http://hackage.haskell.org/package/arithmoi-0.11.0.0/docs/Math-NumberTheory-Primes.html#v:factorBack)?

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -14,173 +14,222 @@ import Data.Semigroup
...
+    valuesOfB = trace ("Ratio: " ++ show ((fromInteger (a * fromIntegral m) :: Double) / (fromInteger (integerSquareRoot (2*n))))) $ filter (<= a `div` 2) $ sqrtsModFactorisation n decompositionOfA
+    valuesOfC = map (\x -> (x * x - n) `div` a) valuesOfB
+    mappingFunctions = zipWith (curry (\ (b, c) x -> a * x * x + 2 * b * x + c)) valuesOfB valuesOfC
```
Bodigrim: Probably does not make much sense to use `curry` - just pattern-match on a nested tuple `((b, c), x)`.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -14,173 +14,222 @@ import Data.Semigroup
...
+  | otherwise                    = findFactor n otherSquares
   where
     factor = gcd (x - y) n
 
 -- This routine outputs an infinite list of tuples @(x, y)@ such that
 -- @x ^ 2 - y ^ 2 `mod` n = 0@. A factorisation can be infered from this data
 -- in at least a half of the cases.
-findSquares :: Integer -> Int -> Int -> [(Integer, Integer)]
-findSquares n b t = runST $ do
+findSquares :: Integer -> Int -> Int -> Int -> [(Integer, Integer)]
```
Bodigrim: Again, pass `QuadraticSieveConfig`.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -14,173 +14,222 @@ import Data.Semigroup
...
+    higherPrimes = take (len - length lowerPrimes) $ filter positiveResidue $ generatePrimesForwards $ floor midPoint + 1
+    -- The length of @lowerPrimes@ may not be @len `div` 2@
+    lowerPrimes = take (len `div` 2) $ filter positiveResidue $ generatePrimesBackwards $ floor midPoint
+    positiveResidue p = jacobi n ((intToInteger . unPrime) p) == One
+
+generatePrimesForwards :: Int -> [Prime Int]
+generatePrimesForwards from
+  -- 2 cannot be a factor of a
+  | from <= 2 = generatePrimesForwards 3
+  | otherwise = [nextPrime from..]
```
Bodigrim: I'd rather define `generatePrimeForwards from = [nextPrime (max 3 from) ..]`

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -14,173 +14,222 @@ import Data.Semigroup
...
-  where
-    go (onlyOnce, atLeastOnce) x =
-      ((onlyOnce PS.\\ PS.unPrimeIntSet x) <> (x PS.\\ PS.unPrimeIntSet atLeastOnce), atLeastOnce <> x)
+    -- Adjust fudge factor to taste
+    selectSmooth index (value, logResidue) = case logResidue < 18 of
+      True
+        -- Cannot use factorisation instead of listFactorisation since its rightmost value need not be an int.
+        | null listFactorisation || (fst . last) listFactorisation <= intToInteger t -> Just (a * intToInteger (index - m) + b, factorisation)
+        | otherwise                                                                  -> Nothing
+      False                                                                          -> Nothing
```
Bodigrim: Do not take alignment to such extremes. Better put `Just` / `Nothing` on a next line.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -14,173 +14,222 @@ import Data.Semigroup
...
+
+-- Find all primes, which appear only once in the input list.
+appearsOnlyOnce :: [S.IntSet] -> S.IntSet
+appearsOnlyOnce = fst . L.foldl' go (mempty, mempty)
+  where
+    go (onlyOnce, atLeastOnce) x =
+      ((onlyOnce S.\\ x) <> (x S.\\ atLeastOnce), atLeastOnce <> x)
+
+convertToSet :: I.IntMap Int -> S.IntSet
+convertToSet = I.foldrWithKey (\key pow set -> if odd pow then key `S.insert` set else set) mempty
```
Bodigrim: I'd rather filter `IntMap` by condition and use `I.keysSet` next.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -212,19 +259,22 @@ convertToList (SomeKnown solution) = listBits $ SU.fromSized $ unDBVector soluti
 
 -- Given a solution, it computes the product of the numbers in the first
 -- component of the tuple.
-findFirstSquare :: Integer -> V.Vector Integer -> [Int] -> Integer
-findFirstSquare n values = foldr construct 1
-  where
-    construct solutionIndex previous = ((values V.! solutionIndex) * previous) `mod` n
+-- Consider using Mod for clarity.
```
Bodigrim: This is a good idea, let's return `Mod n` instead of `Integer` here and in `findSecondSquare`.

Bodigrim (commented):
Please do not forget to clean up `trace`s and fix Travis build.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
federico-bongiorno: Most of the line is tracing. It is going to go eventually. Should I still split it?

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
federico-bongiorno: It's much better

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
Bodigrim: Probably no.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
+  where
+    -- + 4 for large prime variation
+    h = intLog2 t + 1
+    k
+      | l <= 20   = 1
+      | l <= 30   = 2
+      | l <= 40   = 3
+      | l <= 50   = 4
+      | l <= 60   = 5
+      | otherwise = 6
```
Bodigrim: What about ``k = max 1 (min 6 (l `div` 10))``?

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
+    m = 3 * t `div` 2
+    t
+      | l <= 5    = floor (exp . sqrt $ log (fromInteger n) * log (log (fromInteger n)) :: Double)
+      | l <= 8    = integerToInt $ integerSquareRoot n
+      | l <= 20   = floor ((*32) . sqrt . exp . sqrt $ log (fromInteger n) * log (log (fromInteger n)) :: Double)
+      | l <= 25   = floor ((*25) . sqrt . exp . sqrt $ log (fromInteger n) * log (log (fromInteger n)) :: Double)
+      | l <= 30   = floor ((*20) . sqrt . exp . sqrt $ log (fromInteger n) * log (log (fromInteger n)) :: Double)
+      | l <= 35   = floor ((*15) . sqrt . exp . sqrt $ log (fromInteger n) * log (log (fromInteger n)) :: Double)
+      | l <= 40   = floor ((*10) . sqrt . exp . sqrt $ log (fromInteger n) * log (log (fromInteger n)) :: Double)
+      | otherwise = floor ((*5) . sqrt . exp . sqrt $ log (fromInteger n) * log (log (fromInteger n)) :: Double)
```
Bodigrim: Let's avoid repetitions, please define a helper function.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
--- | Given an odd positive composite Integer @n@ and Int parameters @b@ and @t@,
--- the Quadratic Sieve attempts to output @factor@, a factor of @n@. The
+data QuadraticSieveConfig = QuadraticSieveConfig
+  { qscFactorBase :: Int
+  , qscSievingInterval :: Int
+  , qscNumberOfBlocks :: Int
+  , qscSmoothThreshold :: Int
+  }
+
+autoConfig :: Integer -> QuadraticSieveConfig
```
Bodigrim, line 86: Nice!

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
     factor = gcd (x - y) n
 
 -- This routine outputs an infinite list of tuples @(x, y)@ such that
 -- @x ^ 2 - y ^ 2 `mod` n = 0@. A factorisation can be infered from this data
 -- in at least a half of the cases.
-findSquares :: Integer -> Int -> Int -> [(Integer, Integer)]
-findSquares n b t = runST $ do
+findSquares :: Integer -> QuadraticSieveConfig -> [(Integer, Integer)]
+-- k indicates maximum number of sieve blocks to go thorugh 2 ^ (k - 1)
+findSquares n qsc = runST $ do
```
Bodigrim: One can save some indirections, pattern-matching by `QuadraticSieveConfig t m k h` immediately.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
-    squareRoot = integerSquareRoot n
-    mappingFunction j = j * j - n
-    startingPoint = squareRoot - intToInteger t `div` 2
+    t = qscFactorBase qsc
+    m = qscSievingInterval qsc
+    k = qscNumberOfBlocks qsc
+    h = qscSmoothThreshold qsc
+    factorBase = [nextPrime 2..precPrime t]
+    -- Make sure rootOfA is an Int
+    rootOfA = ((2 * fromInteger n) ** (1 / (4 * fromIntegral k))) / (fromIntegral m ** (1 / (2 * fromIntegral k))) :: Double
```
Bodigrim: It should be fine for the expected range of `n`, `m` and `k`, but to be on the safe side let's stick to integer arithmetic and use [`integerRoot`](https://hackage.haskell.org/package/integer-roots-1.0/docs/Math-NumberTheory-Roots.html#v:integerRoot) to take 4k-th root of (2n/m^2).

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
+    -- Classical quadratic sieve if rootOfA < 5
+    -- listOfFactors = constructA n k
```
Bodigrim: What does this comment mean?

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
+    listOfFactors = trace ("Root of a: " ++ show rootOfA) $ if rootOfA < 5 then [] else generatePrimes n rootOfA k
+    initialDecompositionOfA = zip (map (fromJust . toPrimeIntegral) listOfFactors) (repeat 2)
```
Bodigrim: Instead of `zip xs (repeat 2)` one can `map (,2)`. Requires enabling `{-# LANGUAGE TupleSections #-}`.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
-          | numberOfPrimes + 1 < length mat = suitableSmoothNumbers
-          -- If the counter is odd the sieve moves one block to the right.
-          | odd counter                     = goSieving smoothNumbers (newStartingPoint + intToInteger (counter * t)) (counter + 1)
-          -- If the counter is even the sieve moves one block to the left.
-          | otherwise                       = goSieving smoothNumbers (newStartingPoint - intToInteger (counter * t)) (counter + 1)
+    goSieving :: [(Integer, I.IntMap Int)] -> [(Prime Integer, Word)] -> [(Integer, I.IntMap Int)]
+    goSieving previousDiffSmoothNumbers decompositionOfA = goSelfInitSieving previousDiffSmoothNumbers $ zip valuesOfB valuesOfC
+      where
+        a = factorBack decompositionOfA
+        valuesOfB = trace ("Ratio: " ++ show ((fromInteger (a * fromIntegral m) :: Double) / fromInteger (integerSquareRoot (2*n)))) $ filter (<= a `div` 2) $ sqrtsModFactorisation n decompositionOfA
```
Bodigrim: I think we should leave traces (but set `debug = False`) in production code, so let's split this line into two :)

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
-      pure matrix
+            -- This line has to be changed
+            nextDecompositionOfA = decompositionOfA
+
+        goSelfInitSieving previousSmoothNumbers ((b, c) : otherCoeffs) = runST $ do
+          let
+            f x = a * x * x + 2 * b * x + c
+            (sievingInterval, sievingLogInterval) = V.unzip $ generateLogInterval f m
+          sievingLogIntervalM <- V.thaw sievingLogInterval
+          smoothLogSieveM sievingLogIntervalM (zip factorBase squareRoots) a b c m
```
Bodigrim: You zip lists of equal length, but this is not immediately obvious to the reader. I'd rather define `factorBaseWithSquareRoots = map (\p -> (p, ...)) factorBase` doing everything in one pass, instead of `squareRoots`.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
-            numberOfPrimes = PS.size (foldr (\col acc -> acc <> primeSet col) mempty mat)
-            mat = fmap snd suitableSmoothNumbers
-      pure matrix
+            -- This line has to be changed
+            nextDecompositionOfA = decompositionOfA
+
+        goSelfInitSieving previousSmoothNumbers ((b, c) : otherCoeffs) = runST $ do
+          let
+            f x = a * x * x + 2 * b * x + c
+            (sievingInterval, sievingLogInterval) = V.unzip $ generateLogInterval f m
```
Bodigrim: Generating a vector of pairs is expensive, if you mean to unzip it immediately. Let's probably change `generateLogInterval` to generate a pair of vectors (and make `sievingLogInterval` unboxed).

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
+          sievingLogIntervalM <- V.thaw sievingLogInterval
```
Bodigrim: Could you possibly use `unsafeThaw` here? Is original `sievingLogInterval` used by someone else?

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
+          smoothLogSieveM sievingLogIntervalM (zip factorBase squareRoots) a b c m
+          sievedLogInterval <- V.unsafeFreeze sievingLogIntervalM
+          let
+            newSmoothNumbers = V.toList $ findLogSmoothNumbers factorBase m h decompositionOfA b $ V.zip sievingInterval sievedLogInterval
+            smoothNumbers = SS.toList . SS.fromList $ previousSmoothNumbers ++ newSmoothNumbers
```
Bodigrim: This might be expensive. Let's probably make `previousSmoothNumbers, newSmoothNumbers, smoothNumbers :: SS.Set (Integer, I.IntMap Int)` instead of lists.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
+            matrixSmoothNumbers
+              -- Minimise length of matrix
+              | numberOfConstraints < length mat = trace ("Matrix dimension: " ++ show (numberOfConstraints, length mat)) $ take (numberOfConstraints + 10) smoothNumbers
+              | otherwise                        = trace ("Matrix dimension: " ++ show (numberOfConstraints, length mat)) $ goSelfInitSieving smoothNumbers otherCoeffs
```
Bodigrim: You can factor out tracing:
```haskell
matrixSmoothNumbers 
  | trace ... False = undefined 
  | numberOfContraints < length mat = ... 
  | otherwise = ...
```

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
+              where
+                numberOfConstraints = S.size $ foldr (\col acc -> acc <> I.keysSet col) mempty mat
```
Bodigrim: Isn't it `foldMap I.keysSet`?

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
-    goSolving :: Int -> [(Integer, SignedPrimeIntSet)] -> [(Integer, Integer)]
-    goSolving seed sievingData = firstSquare `seq` secondSquare `seq` (firstSquare, secondSquare) : goSolving (seed + 1) sievingData
+    goSolving :: Int -> [(Integer, Integer)]
+    goSolving seed = firstSquare `seq` secondSquare `seq` (firstSquare, secondSquare) : goSolving (seed + 1)
       where
-        firstSquare = findFirstSquare n (V.fromList (fmap fst sievingData)) solution
-        secondSquare = findSecondSquare n (V.fromList (fmap (snd . second primeSet) sievingData)) solution
-        solution = convertToList $ linearSolve' seed $ translate $ fmap snd sievingData
+        firstSquare = findFirstSquare n (fmap fst squaresData)
+        secondSquare = findSecondSquare n (fmap snd squaresData)
```
Bodigrim, line 228: Use `Data.List.unzip` to unzip `squaresData` in one pass.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
+        -- Add factorisation of a
+        squaresData = map (sievingData !!) solution
+        solution = convertToList $ linearSolve' seed matrix
 
-  pure $ goSolving (integerToInt n) indexedSmoothNumbers
+  pure $ goSolving (integerToInt n)
+
+generatePrimes :: Integer -> Double -> Int -> [Prime Int]
+generatePrimes n midPoint len = lowerPrimes ++ higherPrimes
```
Bodigrim: There is an simpler way to do it, if we can sacrifice some precision. Use [`approxPrimeCount`](http://hackage.haskell.org/package/arithmoi-0.11.0.0/docs/Math-NumberTheory-Primes-Counting.html#v:approxPrimeCount) and [`nthPrimeApprox`](http://hackage.haskell.org/package/arithmoi-0.11.0.0/docs/Math-NumberTheory-Primes-Counting.html#v:nthPrimeApprox) to estimate the lower bound (assuming that half of primes are positive residues). Then generate `len` primes starting from this lower bound.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
-findSmoothNumbers startingPoint = V.imapMaybe selectSmooth
+findLogSmoothNumbers :: [Prime Int] -> Int -> Int -> [(Prime Integer, Word)] -> Integer -> V.Vector (Integer, Int) -> V.Vector (Integer, I.IntMap Int)
+findLogSmoothNumbers factorBase m h decompositionOfA b = V.imapMaybe selectSmooth
   where
-    selectSmooth index (residue, factorisation)
-      | residue == 1 = Just (intToInteger index + startingPoint, factorisation)
-      | otherwise    = Nothing
+    a = factorBack decompositionOfA
+    intMapA = I.fromAscList $ map (bimap (integerToInt . unPrime) wordToInt) decompositionOfA
+    highestPrime = intToInteger . unPrime . last $ factorBase
```
Bodigrim: `factorBase` is sorted, but since types itself does not guarantee it, I'd probably use `maximum` instead of `last` here.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
+    selectSmooth index (value, logResidue) = case logResidue < h of
+      True
+        -- Cannot use factorisation instead of listFactorisation since its rightmost value need not be an int.
+        | null listFactorisation                         -> Just (a * intToInteger (index - m) + b, factorisation)
+        | (fst . last) listFactorisation <= highestPrime -> Just (a * intToInteger (index - m) + b, factorisation)
```
Bodigrim: What's the reason to duplicate right-hand side?

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,289 @@
...
+        | otherwise                                      -> Nothing
+      False -> Nothing
+      where
+        factorisation = I.unionWith (+) intMapA $ if value < 0 then I.insert (-1) 1 preFac else preFac
+        preFac = I.fromAscList $ map (bimap integerToInt wordToInt) listFactorisation
+        -- Maybe there is a better way to use @trialDivision@
+        listFactorisation = trialDivisionWith (map (intToInteger . unPrime )factorBase) value
```
Bodigrim: Is `trialDivisionWith` taking a significant time? There are ways to speed up it a bit.

app/QuadraticSieve.hs
```diff
@@ -1,8 +1,22 @@
...
+-- main :: IO ()
+-- main = print $ quadraticSieve n t (2 * t) 3
+--   where
+--     t = floor b
+--     b = (*15) . sqrt . exp . sqrt $ log (fromInteger n) * log (log (fromInteger n)) :: Double
+--     n = 313171213163547737474104867123
+--       -- 313171213163547737474104867123
+--       -- 820239374043355485537790987979115484333
+--       -- 1000000000100000000000000000074000000005700000000000000000969
+
```
Bodigrim, line 3: Remove commented code?

test-suite/Math/NumberTheory/Primes/QuadraticSieveTests.hs
```diff
@@ -7,21 +7,38 @@ import Test.Tasty.QuickCheck as QC
...
+--   -- The value of @b@ is too low for sieving to be successful.
+--   | n < 100000                 = True
+--   | otherwise                  = (firstSquare ^ (2 :: Int) - secondSquare ^ (2 :: Int)) `mod` n == 0
+--     where
+--       (firstSquare, secondSquare) = trace ("Number: " ++ show n) $ head $ findSquares n t (2 * t) 1
+--       t = floor l
+--       l = (*25) . sqrt . exp . sqrt $ log (fromInteger n) * log (log (fromInteger n)) :: Double
+--       n = toInteger p * toInteger q
+--       p = unPrime . nextPrime $ i `mod` 100000000
+--       q = unPrime . nextPrime $ j `mod` 100000000
```
Bodigrim: Remove commented code?

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
federico-bongiorno: It makes sense, I left it this way so I could tweak it later. These are just provisional values

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
federico-bongiorno: Is there a way for maximum to return a default value on an empty list?

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
federico-bongiorno: It can take some time if the smoothness threshold is high. I think it's worth trying to speed it up

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
Bodigrim: Could you possibly just use `qscFactorBase`?

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
federico-bongiorno: I haven't changed this yet because I am not sure it is what I want. We can discuss this later.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -55,6 +56,9 @@ instance KnownNat k => Monoid (DBVector k) where
   mempty = DBVector $ SU.replicate (Bit False)
   mappend = (<>)
 
+intVal :: KnownNat k => a k -> Int
+intVal = naturalToInt . natVal
```
Bodigrim: Let's probably move this helper to `M.NT.Utils`.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,277 @@
...
+autoConfig n = QuadraticSieveConfig t m k h
+  where
+    -- + 4 for large prime variation
+    h = intLog2 t + 1
+    k = max 1 (l `div` 10)
+    m = 3 * t `div` 2
+    t
+      | l < 4    = integerToInt n `div` 2
+      | l < 8    = integerToInt $ integerSquareRoot n
+      | otherwise = (50 - l) * floor (sqrt . exp . sqrt $ log (fromInteger n) * log (log (fromInteger n)) :: Double)
```
Bodigrim: Again, let's avoid converting huge integers to `Double`. Approximate this calculation using [`integerLog2`](http://hackage.haskell.org/package/integer-logarithms-1.0.3/docs/Math-NumberTheory-Logarithms.html#v:integerLog2).

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,277 @@
...
   where
     factor = gcd (x - y) n
 
 -- This routine outputs an infinite list of tuples @(x, y)@ such that
 -- @x ^ 2 - y ^ 2 `mod` n = 0@. A factorisation can be infered from this data
 -- in at least a half of the cases.
-findSquares :: Integer -> Int -> Int -> [(Integer, Integer)]
-findSquares n b t = runST $ do
+findSquares :: Integer -> QuadraticSieveConfig -> [(Integer, Integer)]
+-- k indicates maximum number of sieve blocks to go thorugh 2 ^ (k - 1)
```
Bodigrim: "thorugh" -> "through"

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,277 @@
...
+findSquares n (QuadraticSieveConfig t m k h) = runST $ do
   let
-    factorBase = [nextPrime 2..precPrime b]
-    squareRoot = integerSquareRoot n
-    mappingFunction j = j * j - n
-    startingPoint = squareRoot - intToInteger t `div` 2
+    factorBase = filter (\p -> unPrime p == 2 || jacobi n ((intToInteger . unPrime) p) == One) [nextPrime 2..precPrime t]
```
Bodigrim: Let's define `isResidue :: Integer -> Prime Int -> Bool` as a separate helper function and use it here and below.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -5,182 +5,277 @@
...
+    -- Make sure rootOfA is an Int
+    rootOfA = floor (((2 * fromInteger n) ** (1 / (4 * fromIntegral k))) / (fromIntegral m ** (1 / (2 * fromIntegral k))) :: Double)
+    -- Classical quadratic sieve if rootOfA < 5
+    -- Replace with better way to check whether all primes are less than the largest Prime Int
+    factorsOfA = if rootOfA < 5 then [] else generatePrimes n rootOfA k
+    initialDecompositionOfA = trace ("Root of a: " ++ show rootOfA) $ zip (if all ((< 9223372036854775784) . unPrime) factorsOfA then factorsOfA else error "Parameters are not large enough.") (repeat 2)
```
Bodigrim: Do not use magic numbers. Something like `toInteger (maxBound :: Int)` will be better.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -24,28 +25,32 @@ import qualified Data.Vector.Unboxed.Sized as SU
...
 newtype DBVector (k :: Nat) = DBVector { unDBVector :: SU.Vector k Bit }
-  deriving (Eq, Show)
+  deriving (Eq, Ord, Show)
 
 -- | Sparse Binary square Matrix of size @k@. It is formed of columns
 -- of sparse binary vectors.
 newtype SBMatrix (k :: Nat) = SBMatrix { unSBMatrix :: SV.Vector k (SBVector k) }
 
+instance KnownNat k => Show (SBVector k) where
+  show (SBVector sbVec) = show $ U.map (wordToInt . MW.unMod) sbVec
```
Bodigrim: `wordToInt` does not make any difference for the purposes of `show`.

Math/NumberTheory/Primes/Factorisation/LinearAlgebra.hs
```diff
@@ -55,6 +60,9 @@ instance KnownNat k => Monoid (DBVector k) where
   mempty = DBVector $ SU.replicate (Bit False)
   mappend = (<>)
 
+instance KnownNat k => Show (SBMatrix k) where
+  show (SBMatrix mat) = show $ SV.toList mat
```
Bodigrim: Is it better than an auto-derived one?

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -1,186 +1,292 @@
...
-    factorBase = [nextPrime 2..precPrime b]
-    squareRoot = integerSquareRoot n
-    mappingFunction j = j * j - n
-    startingPoint = squareRoot - intToInteger t `div` 2
+    factorBase = filter (isResidue n) [nextPrime 2..precPrime t]
+    rootOfA
+      | k <= 0    = 1
+      | otherwise = integerRoot (4 * k) ((2 * n) `div` (intToInteger m ^ (2 :: Int)))
+    factorsOfA = fromMaybe (error "Parameters are not large enough.") $ mapM isPrimeInt $ generatePrimes n rootOfA k
+    initialDecompositionOfA = trace ("Root of a: " ++ show rootOfA) $ zip factorsOfA (repeat 2)
```
Bodigrim: `map (,2) factorsOfA`, please.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -1,186 +1,292 @@
...
+
+        goSelfInitSieving :: [(Integer, I.IntMap Int)] -> [(Integer, Integer)] -> [(Integer, I.IntMap Int)]
+        goSelfInitSieving previousSmoothNumbers [] = goSieving previousSmoothNumbers nextDecompositionOfA
           where
-            numberOfPrimes = PS.size (foldr (\col acc -> acc <> primeSet col) mempty mat)
-            mat = fmap snd suitableSmoothNumbers
-      pure matrix
+            -- Must check that int does not overflow
+            nextDecompositionOfA
+              | null decompositionOfA = zip (generatePrimes n 1 1) [2]
```
Bodigrim: What does it mean? It looks really suspicious to zip with `[2]`.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
Bodigrim: Or even `Map Integer (I.IntMap Int)`.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -1,186 +1,292 @@
...
+  | otherwise = precPrime to : generatePrimesBackwards (unPrime (precPrime to) - 1)
+
+isPrimeInt :: Prime Integer -> Maybe (Prime Integer)
+isPrimeInt x = fromJust . toPrimeIntegral <$> primeInt
+  where
+    primeInt = toPrimeIntegral x :: Maybe (Prime Int)
 
-  pure $ goSolving (integerToInt n) indexedSmoothNumbers
+isResidue :: Integer -> Prime Int -> Bool
+isResidue n p = (unPrime p == 2) || jacobi n ((intToInteger . unPrime) p) == One
```
Bodigrim: We can do a bit better here. Instead of calling `jacobi` with two `Integer` arguments, we can reduce `n` modulo `p`, cast to `Int` and call `jacobi` for two `Int`s, which should be faster.

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -1,186 +1,292 @@
...
+generatePrimesForwards from = [nextPrime (max 3 from)..]
+
+generatePrimesBackwards :: Integer -> [Prime Integer]
+generatePrimesBackwards to
+  -- 2 cannot be a factor of a
+  | to <= 2   = []
+  | otherwise = precPrime to : generatePrimesBackwards (unPrime (precPrime to) - 1)
+
+isPrimeInt :: Prime Integer -> Maybe (Prime Integer)
+isPrimeInt x = fromJust . toPrimeIntegral <$> primeInt
```
Bodigrim: If `primeInt` is not `Nothing` you can simply return `x`, avoiding extra operations: `primeInt >> pure x`.

test-suite/Math/NumberTheory/Primes/QuadraticSieveTests.hs
```diff
@@ -7,23 +7,36 @@ import Test.Tasty.QuickCheck as QC
...
+
+checkSquares :: Large Int -> Large Int -> Bool
+checkSquares (Large i) (Large j)
+  | p == 2 || q == 2 || p == q = True
+  | otherwise                  = (x * x - y * y) `mod` n == 0
+  where
+    (x, y) = trace ("Number: " ++ show n) $ head $ findSquares n $ autoConfig n
+    n = p * q
+    p = toInteger . unPrime . nextPrime $ i `mod` 100000000
+    q = toInteger . unPrime . nextPrime $ j `mod` 100000000
```
Bodigrim: Is it still desirable to reduce modulo 100000000 here?

federico-bongiorno: Yes, but the tests will take longer

Math/NumberTheory/Primes/Factorisation/QuadraticSieve.hs
```diff
@@ -1,188 +1,355 @@
...
+            Nothing          -> error "Math.NumberTheory.Primes.Factorisation.QuadraticSieve: Algorithm incorrect."
+      forM_ startingIndices $ \startingIndex -> do
+        let change y = y - (intLog2 . unPrime) prime
+        forM_ [startingIndex, startingIndex + unPrime prime..(2 * m)] $ \entry ->
+          MU.modify sievingIntervalM change entry
+
+-- This routine takes the @sievedInterval@ as input, it first filters for smooth numbers
+-- using the given threshold and then checks which of the filtered numbers are smooth
+-- by computing their factorisation by trial division.
+findLogSmoothNumbers :: [Prime Int] -> Int -> Int -> [(Prime Integer, Word)] -> Integer -> V.Vector (Integer, Int) -> V.Vector (Integer, I.IntMap Int)
```
Bodigrim: I suggest changing this function to take `V.Vector Integer` and `U.Vector Int` separately, and return `Map Integer (I.IntMap Int)` immediately.

PR #213, opened 2020-12-27
Fix documentation for construction of MultMod m

cjlarose:
There's no function called `toMultElement` as far as I can tell.


b-mehta (approved):

Bodigrim:
Thanks!

PR #215, opened 2021-02-14
Clarify seemingly broken links to combinat functions in haddock

jhrcek:
At first I was confused because the link in haddock to `stirling2nd` in the [haddock of stirling2](https://hackage.haskell.org/package/arithmoi-0.11.0.1/docs/Math-NumberTheory-Recurrences-Bilinear.html#v:stirling2) seemed broken. Searching for `stirling2nd` in the index of arithmoi led me to conclude it was removed from this package in the past.
Then I hoogled for `stirling2nd` and it didn't return any results. Later I found this probably refers to `stirling2nd` from the `combinat` package, but hoogle doesn't find it because it searches only packages which are on stackage (and `combinat` is not in recent stackage snapshots). So I figured it might be useful to link to the package from all the (3) places mentioning some function from `Math.Combinat...` namespace.


Bodigrim:
Thanks!

PR #220, opened 2022-11-15
linear diophantine equations

BlackCapCoder:
These come up quite a bit, and I've looked for them in arithmoi more than once!

`Kuṭṭaka`, `Pogo` and `Race` are usage examples if you will, rather than something I want into the library. `Pogo` in particular is pretty interesting: any Brainfuck program that does not contain nested loops can be reduced to a Pogo problem!

### DIFF
--- a/Math/NumberTheory/Diophantine.hs
+++ b/Math/NumberTheory/Diophantine.hs
@@ -1,6 +1,6 @@
 -- Module for Diophantine Equations and related functions
 
-{-# LANGUAGE RecordWildCards, PartialTypeSignatures #-}
+{-# LANGUAGE PartialTypeSignatures #-}
 {-# OPTIONS_GHC -Wno-partial-type-signatures #-}
 
 module Math.NumberTheory.Diophantine
@@ -77,7 +77,7 @@ cornacchia d m
 ----
 
 -- | A solution to a linear equation
-data LinearSolution a = LS { x,v,y,u :: a }
+data LinearSolution a = LS { base1,base2,scale1,scale2 :: a }
   deriving
     ( Show, Eq, Ord
     )
@@ -87,7 +87,7 @@ data LinearSolution a = LS { x,v,y,u :: a }
 -- | where `x` and `y` are unknown
 linear :: _ => a -> a -> a -> Maybe (LinearSolution a)
 linear a b c =
-    LS {..} <$ guard (b /= 0 && q == 0)
+    LS x y v u <$ guard (b /= 0 && q == 0)
   where
     (d, e) = gcdExt a b
     (h, q) = divMod c d
@@ -98,6 +98,6 @@ linear a b c =
 -- | Produces an unique solution given any
 -- | arbitrary number k
 runLinearSolution :: _ => LinearSolution a -> a -> (a, a)
-runLinearSolution LS {..} k =
+runLinearSolution (LS x y v u) k =
   ( x + k*v, y - k*u )
 
--- a/Math/NumberTheory/Diophantine.hs
+++ b/Math/NumberTheory/Diophantine.hs
@@ -1,8 +1,16 @@
 -- Module for Diophantine Equations and related functions
 
+{-# LANGUAGE RecordWildCards, BlockArguments, TypeApplications, PartialTypeSignatures #-}
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# OPTIONS_GHC -Wno-partial-type-signatures #-}
+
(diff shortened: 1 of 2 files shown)


Math/NumberTheory/Diophantine.hs
```diff
@@ -64,3 +75,101 @@ cornacchia d m
...
+
+-- | A solution to a linear equation
+data LinearSolution a = LS { x,v,y,u :: a }
+  deriving
+    ( Show, Eq, Ord
+    )
+
+-- | Produces an unique solution given any
+-- | arbitrary number k
+runLinearSolution :: _ => LinearSolution b -> b -> (b, b)
```
Bodigrim: Could you please add tests, showing that 
a) everything this function returns **is** a solution,
b) all possible solutions are generated by this function by supplying different `k`?

BlackCapCoder: There are either no solutions (handled by Maybe), or an infinite number of solutions. I'll add some test cases, but I cannot check it exhaustively

test-suite/Math/NumberTheory/DiophantineTests.hs
```diff
@@ -33,8 +34,17 @@ cornacchiaBruteForce (Positive d) (Positive a) = gcd d m /= 1 || findSolutions [
             where x2 = m - d*y*y
                   x = integerSquareRoot x2
 
+linearTest :: (a ~ Integer) => a -> a -> a -> a -> Bool
+linearTest a b c k =
+  case solveLinear Lin {..} of
+    Nothing -> True -- Disproving this would require a counter example
```
Bodigrim, line 41: The problem is that `solveLinear = const Nothing` passes this test. Can we come up with a test which requires `solveLinear` return `Just{}` at least in some cases?

BlackCapCoder: Yes! I think this is easier to see in terms of my Pogo problem `D = H*L (mod C)` where `H` is the solution. All we have to do is make sure `L` is coprime to `C`, because `sort [mod (h*l) c | h <- [0..pred c]] == [0..pred c]`, which includes `D`.

Math/NumberTheory/Diophantine.hs
```diff
@@ -64,3 +74,35 @@ cornacchia d m
  where
   candidates = map (\sf -> (sf, m `div` (sf * sf))) (squareFactors m)
   solve (sf, m') = map (\(x, y) -> (x * sf, y * sf)) (cornacchiaPrimitive d m')
+
+----
+
+-- | A linear diophantine equation `ax + by = c`
+-- | where `x` and `y` are unknown
+data Linear a = Lin { a,b,c :: a }
```
Bodigrim: This makes `a`, `b` and `c` globally defined functions, stealing them from any other potential usage. Could you please use something more specific and less likely to clash? Same applies to `LinearSolution`.

BlackCapCoder: If that is a problem I am more inclined to not have names at all, or something like `a_var`, or removing `Linear` all together in favor of function arguments? The current names are what is used in the literature

BlackCapCoder: I am not as attached to the names in `LinearSolution`. The only reason I have it rather than returning `Maybe (Int -> (Int, Int))` is that `runLinearSolution` is a very simple function, and applying it is technically lossy

BlackCapCoder: Sorry for the spam.

I am leaning towards just removing `Linear`. It is only used in one function- the user can create their own data type if they really want it.

On the other hand, I know that `Linear` has a sensible `Semigroup` instance, and many analogues of `liftA2 f` makes sense for it:
```haskell
-- Basically `liftA2 (+)`
-- (a1+a2)*x + (b1+b2)*y = c1 + c2
--
add (Lin a1 b1 c1) (Lin a2 b2 c2) =
  Lin (a1+a2) (b1+b2) (c1+c2)
```
I didn't add instances because I didn't want to bloat this PR.

Math/NumberTheory/Diophantine.hs
```diff
@@ -1,8 +1,14 @@
 -- Module for Diophantine Equations and related functions
 
+{-# LANGUAGE PartialTypeSignatures #-}
+{-# OPTIONS_GHC -Wno-partial-type-signatures #-}
```
Bodigrim, line 4: Is it difficult to provide full type signatures?

PR #224, opened 2025-11-27
Add "-O2" to all "ghc_options" parts of cabal file...

GordonBGood:
cabal generally installs packages with default -O1 optmization unless forced by a manual install - the only way to force it always to be installed with optimization is to add this to the cabal file.

Performance testing shows that the extra level of optimization gives a boost in execution speed of about two times, and for the prime counting function about two and a half times.

### DIFF
--- a/arithmoi.cabal
+++ b/arithmoi.cabal
@@ -102,7 +102,7 @@ library
     Math.NumberTheory.Zeta.Riemann
     Math.NumberTheory.Zeta.Utils
   default-language: Haskell2010
-  ghc-options: -Wall -Widentities -Wcompat -Wno-deprecations
+  ghc-options: -Wall -Widentities -Wcompat -Wno-deprecations -O2
 
 test-suite arithmoi-tests
   build-depends:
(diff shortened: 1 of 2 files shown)


Bodigrim:
If there is a noticable speed up from `-O2` in the library component I'm all for it. Thanks!

I'm a bit skeptical about enabling `-O2` in the test suite: my gut feeling is that we'll lose more in compilation time than gain from faster execution of tests. All computationally heavy parts are already covered by `-O2` in the library component, so we just strain ourselves to compile QuickCheck properties and stuff, which is unlikely to bear much fruit.

As for benchmarks, they are supposed to be executed with a timeout, like, `cabal run arithmoi-bench -- --timeout 100`, otherwise they are prone to spin cycles for a long time, waiting for stable results. This is especially pronounced for tests allocating a lot (e. g., triangular tables of long integers). I'm fine with your changes to `benchTriangle`, it should indeed reduce benchmarking time to a reasonable duration.

GordonBGood:
I'm closing this PR and opening another, as I've learned that the better way to do this is by adding a cabal.project file with the specified optimization level in it, so that's what I'll recommend in my next PR.

PR #225, opened 2025-11-28
Turn on optimization level 2; shorten long benchmarks...

GordonBGood:
cabal generally installs packages with default -O1 optmization unless forced by a manual install - the recommended way to force it always to be installed with optimization is to add  a "cabal.project" file with the high optimization specified as here.

Performance testing shows that the extra level of optimization gives a boost in execution speed of about two times, and for the prime counting function about two and a half times.


GordonBGood:
I screwed up and mistakenly PR'ed too many unintended commits together, including a broken new version of the prime count implementation module; I'm closing this PR, reverting the mistaken commits, and will resubmit a new PR with only the optimization improvements...

PR #227, opened 2025-12-12
New legendre primecount

GordonBGood:
- changed primeCount implementation to something a little faster and better documented.
- added a documentation file: "HowPrimeCountingWorks.md".
- upped the version of "integer-roots" to 1.0.4.0 so `Word64` is a specialization for square root.
- made the optimization default to level 2.
- tweaked some benchmark settings to make them complete faster.

### DIFF
--- a/Math/NumberTheory/Primes/Counting/HowPrimeCountingWorks.md
+++ b/Math/NumberTheory/Primes/Counting/HowPrimeCountingWorks.md
@@ -1,0 +1,153 @@
+## Algorithm and Implementation Description
+
+### History and Formulas
+
+This prime counting implementation is based on the work by Adrien-Marie Legendre, who in about the year 1800 observed that the count of the primes to a limit, Pi(limit), can be determined by the inclusion-exclusion principle as noted at the following link:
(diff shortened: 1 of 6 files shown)


Bodigrim:
As a crude benchmark, I ran `cabal run tests -- -p Primes.Counting` before the changes to `primeCount` (but after `-O2`) and after. It used to be 1.65s on my machine, but now takes only 0.54s. I think this is awesome, thanks for your contribution!

GordonBGood:
@Bodigrim,

> I think this is awesome, thanks for your contribution!...

I'm happy to help, and yes, that is about the expected ratio of improved performance on top of the about 2.5 times due to using `-O2` optmization.

As I have mentioned before, the Meissel algorithm I am working on won't be all that much faster than this Legendre implementation, at least for smallish counting ranges such as this, but of course the main reason for wanting it is to reduce the memory consumption from O(n^(1/2)) to O(n^(1/3)).  There is also some room for some constant factor speed improvements by improving the sieving, but that gets more and more complex...

PR #231, opened 2026-05-31
`ids` and `associates` for `GaussianIntegers`

anshwad10:
.


Bodigrim:
Thanks! (I wonder if we should rename `ids` to `units` in both modules)

